PR #2, opened 2016-08-14
use ThreadLocal in SampleConsumer because SimpleDateFormat is thread unsafe.

toastkidjp:
Nice to meet you. It's my first pull request.
I modified to use ThreadLocal because SimpleDateFormat is thread unsafe.
Please could you check it when you get time? :bow:


rajinisivaram:
@toastkidjp @akozich Thank you for the PR. The samples use `SimpleDateFormat` only in a single thread.

toastkidjp:
@akozich @rajinisivaram 
Thank you for your reply. I close it.

PR #3, opened 2016-08-14
modify dependencies gs-collections to Eclipse Collections 7.1.0.

toastkidjp:
GS-Collections upgraded to [Eclipse Collections](https://github.com/eclipse/eclipse-collections). I modified its dependencies. But, I think it's no longer use.
Please would you check this PR when you get time? :bow:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ ext {
     joptVersion = '4.9'
 
     assertjVersion = '3.3.0'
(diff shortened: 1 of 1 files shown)


rajinisivaram:
@toastkidjp Thank you for the PR. This dependency is no longer used.

toastkidjp:
@rajinisivaram I see . Thank you for your reply. I'll close it

PR #5, opened 2016-12-23
switch to reactor-core and addons 3.0.4

simonbasle:
@rajinisivaram `reactor-core` and `reactor-test` 3.0.4 have been released.
This PR also takes a change in `expectNextCount` behavior in reactor-test 3.0.4 into account.


rajinisivaram:
@simonbasle Thank you for the PR.

PR #12, opened 2017-09-05
Adapt to core Operators hooks now requiring a Context

simonbasle:
The DefaultKafkaSender passes its actual's Context for that purpose.


simonbasle:
@ilayaperumalg I still merged this one despite the regression test failing on one test build, so that the project compiles. Locally it passes, but this may warrant further investigation.

ilayaperumalg:
thanks @simonbasle Created https://github.com/reactor/reactor-kafka/issues/13 for investigating the test issue. I don't seem to find the actual build that reported the test failure. Could you point me to the same?

PR #27, opened 2018-03-27
Deserializers arguments for KafkaReceiver

mvh77:
I'd need to be able to give key and value deserializers directly as arguments to the KafkaReceiver.create method instead of the ReceiverOptions. This is because something like Protobuf generated classes have a static parseFrom method to deserialize objects for which I cannot create a generic deserializer. Instead I'd want to create the deserializers through a factory like the one below:

```
import io.vavr.CheckedFunction1;
import org.apache.kafka.common.serialization.Deserializer;

import java.util.Map;

public class ProtobufDeserializerFactory {

    public static <T> Deserializer<T> of(CheckedFunction1<byte[], T> f) {
        return new Deserializer<T>() {
            @Override
            public void configure(Map<String, ?> configs, boolean isKey) {}

            @Override
            public T deserialize(String topic, byte[] data) {
                try {
                    return f.apply(data);
                } catch (Throwable throwable) {
                    throw new RuntimeException(throwable);
                }
            }

            @Override
            public void close() {}
        };
    }
}
```
and use them a so:

```
KafkaReceiver
    .create(options, 
            new StringDeserializer(), 
            ProtobufDeserializerFactory.of(Definitions.ExampleMessage::parseFrom))
    .receive();
```
In my case this is not required for serializers although for symmetry reasons this could also be implemented in KafkaSender.

### DIFF
--- a/src/main/java/reactor/kafka/receiver/KafkaReceiver.java
+++ b/src/main/java/reactor/kafka/receiver/KafkaReceiver.java
@@ -21,6 +21,7 @@ import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 
+import org.apache.kafka.common.serialization.Deserializer;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.kafka.receiver.internals.ConsumerFactory;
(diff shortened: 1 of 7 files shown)


pivotal-issuemaster:
@mvh77 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-kafka&pullRequestId=27)!

[Click here](https://cla.pivotal.io/sync/reactor?repositoryId=reactor/reactor-kafka&pullRequestId=27) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@mvh77 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-kafka&pullRequestId=27)!

smaldini:
Added in #32

Issue #28, opened 2018-04-12
Reactor Kafka doesn't work with latest reactor-core 3.1.6


rgeorgiev:
The new reactor-core has a logic to validate blocking calls on parallel() and single() Schedulers, this leads to the below exception:
java.lang.IllegalStateException: block()/blockFirst()/blockLast() are blocking, which is not supported in thread parallel-3
	at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:77)
	at reactor.core.publisher.Mono.block(Mono.java:1175)
	at reactor.kafka.receiver.internals.DefaultKafkaReceiver.lambda$receiveAtmostOnce$5(DefaultKafkaReceiver.java:176)

mayras:
Yes, same issue! Returned to 3.1.5 and it works

smaldini:
This is a major issue to be fixed asap

OlegDokuka:
Fixed by #35. Now should be available in the 1.0.1.BUILD-SNAPSHOT

PR #35, opened 2018-06-25
#28

OlegDokuka:
- removed `.block` from the usage by replacing with proper `concatMap()`
- update gradle to the latest version
- update Reactor to 3.1.8
- fix gitignore

### DIFF
--- a/reactor-kafka-samples/src/main/java/reactor/kafka/samples/SampleScenarios.java
+++ b/reactor-kafka-samples/src/main/java/reactor/kafka/samples/SampleScenarios.java
@@ -110,12 +110,12 @@ public class SampleScenarios {
             Flux<Person> srcFlux = source().flux();
             return sender(senderOptions)
                     .send(srcFlux.map(p -> SenderRecord.create(new ProducerRecord<>(topic, p.id(), p), p.id())))
-                    .doOnError(e-> log.error("Send failed, terminating.", e))
+                    .doOnError(e -> log.error("Send failed, terminating.", e))
                     .doOnNext(r -> {
-                            int id = r.correlationMetadata();
-                            log.trace("Successfully stored person with id {} in Kafka", id);
-                            source.commit(id);
-                        })
+                        int id = r.correlationMetadata();
+                        log.trace("Successfully stored person with id {} in Kafka", id);
+                        source.commit(id);
+                    })
                     .doOnCancel(() -> close());
         }
     }
@@ -150,7 +150,7 @@ public class SampleScenarios {
                           .send(Mono.just(new ProducerRecord<>(topic2, p.id(), p.upperCase())))
                           .then()
                           .doOnSuccess(v -> source.commit(p.id())))
-                    .doOnCancel(() -> close());
+                          .doOnCancel(() -> close());
         }
     }
 
@@ -174,7 +174,7 @@ public class SampleScenarios {
                            .receive()
                            .publishOn(scheduler)
                            .concatMap(m -> storeInDB(m.value())
-                                          .doOnSuccess(r -> m.receiverOffset().commit().block()))
+                                          .thenEmpty(m.receiverOffset().commit()))
                            .retry()
                            .doOnCancel(() -> close());
         }
@@ -289,7 +289,7 @@ public class SampleScenarios {
                     .sendTransactionally(srcFlux.map(p -> records(p)))
                     .concatMap(r -> r)
                     .doOnNext(r -> log.info("Sent record successfully {}", r))
-                    .doOnError(e-> log.error("Send failed, terminating.", e))
+                    .doOnError(e -> log.error("Send failed, terminating.", e))
                     .doOnCancel(() -> close());
         }
         private Flux<SenderRecord<Integer, Person, Integer>> records(Person p) {
@@ -398,9 +398,10 @@ public class SampleScenarios {
             return Flux.merge(stream1, stream2)
                        .doOnSubscribe(s -> cancelRef.set(inFlux.subscribe()))
                        .doOnCancel(() -> {
-                               cancel.accept(cancelRef);
-                               close();
-                           });
+                           cancel.accept(cancelRef);
+                           close();
+                       })
+                       .doOnTerminate(() -> close());
         }
         public ProducerRecord<Integer, Person> process1(Person p, boolean debug) {
             if (debug)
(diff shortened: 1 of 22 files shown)


src/test/java/reactor/kafka/receiver/KafkaReceiverTest.java
```diff
@@ -245,7 +245,7 @@ public void manualAssignmentWithCommit() throws Exception {
         Flux<? extends ConsumerRecord<Integer, String>> kafkaFlux =
                 KafkaReceiver.create(receiverOptions)
                         .receive()
-                        .doOnNext(r -> r.receiverOffset().commit().block())
+                        .concatMap(r -> r.receiverOffset().commit().then(Mono.just(r)))
```
smaldini: why not just ``` .concatMap(r -> r.receiverOffset().commit()```

OlegDokuka: Because if we do that we will lose `r` because in `doOnNext` the stream does not map one to another. But concateMap will map `ConsumerRecords` to `Void` so we will lose element.

smaldini: then delayUntil might be the right operation 💃

reactor-kafka-samples/src/main/java/reactor/kafka/samples/SampleScenarios.java
```diff
@@ -174,7 +174,7 @@ public KafkaSource(String bootstrapServers, String topic) {
                            .receive()
                            .publishOn(scheduler)
                            .concatMap(m -> storeInDB(m.value())
-                                          .doOnSuccess(r -> m.receiverOffset().commit().block()))
+                                          .delayUntil(r -> m.receiverOffset().commit()))
```
bsideup: Since `.delayUntil` is implemented as `it -> .concatMap(...).thenReturn(it)`, I would suggest using `.concatMap` here for the same of performance. The result is `Flux<?>`, should be okay

reactor-kafka-samples/src/main/java/reactor/kafka/samples/SampleScenarios.java
```diff
@@ -398,9 +398,9 @@ public FanOut(String bootstrapServers, String sourceTopic, String destTopic1, St
             return Flux.merge(stream1, stream2)
                        .doOnSubscribe(s -> cancelRef.set(inFlux.subscribe()))
                        .doOnCancel(() -> {
-                               cancel.accept(cancelRef);
-                               close();
-                           });
+                           cancel.accept(cancelRef);
+                           close();
```
bsideup: Not related to the indentation change, but shouldn't `close` go to `doFinally`?

src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
```diff
@@ -153,30 +156,37 @@ public DefaultKafkaReceiver(ConsumerFactory consumerFactory, ReceiverOptions<K,
...
         this.ackMode = AckMode.ATMOST_ONCE;
         atmostOnceOffsets = new AtmostOnceOffsets();
-        Flux<ConsumerRecord<K, V>> flux = createConsumerFlux()
-                .concatMap(consumerRecords -> Flux.fromIterable(consumerRecords));
-        return withDoOnRequest(flux)
-                .doOnNext(r -> {
+        return createConsumerFlux()
+            .concatMap(cr -> Flux
+                    .fromIterable(cr)
+                    .concatMap(r -> {
```
bsideup: `.delayUntil` here would help to get rid of `return Mono.just(r)` and `.then(Mono.just(r))` inside the method

src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
```diff
@@ -334,6 +346,11 @@ private void cancel(boolean async) {
                     consumer.wakeup();
                     CloseEvent closeEvent = new CloseEvent(receiverOptions.closeTimeout());
                     if (async) {
+                        if (Thread.currentThread().getName().startsWith("reactive-kafka-")) {
```
bsideup: since this is an important check, I would suggest reusing the constant between this place and the thread creator

src/test/java/reactor/kafka/receiver/KafkaReceiverTest.java
```diff
@@ -441,12 +440,12 @@ public void manualCommitSync() throws Exception {
         receiverOptions = receiverOptions.commitInterval(Duration.ZERO).commitBatchSize(0);
         KafkaReceiver<Integer, String> receiver = createReceiver();
         Flux<? extends ConsumerRecord<Integer, String>> kafkaFlux = receiver.receive()
-                         .doOnNext(record -> {
-                                 assertEquals(committedOffsets[record.partition()], record.offset());
-                                 record.receiverOffset().commit()
-                                       .doOnSuccess(i -> onCommit(record, commitLatch, committedOffsets))
-                                       .block();
-                             })
+                         .concatMap(record -> {
```
bsideup: `.delayUntil` to avoid `.then(Mono.just(record))`?

smaldini:
Minor @bsideup comments should do it, when updated you should go ahead and merge @OlegDokuka

src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
OlegDokuka: removed it for now. Will fix in the #38

src/test/java/reactor/kafka/receiver/KafkaReceiverTest.java
OlegDokuka: done

src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
OlegDokuka: done

reactor-kafka-samples/src/main/java/reactor/kafka/samples/SampleScenarios.java
OlegDokuka: done

reactor-kafka-samples/src/main/java/reactor/kafka/samples/SampleScenarios.java
OlegDokuka: done

Issue #21, opened 2018-02-13
Consumer is not being destroyed when stream terminates


bsideup:
**Problem:**
 if stream terminates normally or because of error, the consumer will not be removed

**Potential reason:**
In `createConsumerFlux` there is `doOnCancel` but no `doOnTerminate`

**Possible solution:**
Replace `doOnCancel` with `doFinally`

PR #38, opened 2018-07-03
#21 fix issue with resource dispose during possible termination

OlegDokuka:


src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
```diff
@@ -322,11 +322,20 @@ private void start() {
...
         log.error("Consumer flux exception", e);
         recordSubmission.error(e);
-        cancel(async);
+        dispose(async);
     }
 
-    private void cancel(boolean async) {
-        log.debug("cancel {}", isActive);
+    private void dispose() {
+        boolean isEventsThread = Thread.currentThread()
```
smaldini: That might be too simplistic - what about keeping the reference in a ThreadLocal that can be populated on start by N schedule or similar.

src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
```diff
@@ -257,7 +256,8 @@ public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
                         if (requestsPending.get() > 0)
                             pollEvent.scheduleIfRequired();
                     })
-                .doOnCancel(() -> cancel(true));
+                .doOnTerminate(this::dispose)
```
bsideup: @smaldini in my code, I often use `doFinally` for this use case.
Is it valid, or `doOnTerminate` + `doOnCancel` is better?

OlegDokuka: Do Finally has different semantic. If first emit signal to downstream and then execute handler where doOnCancel and doOnTerminal exexutes handler first and then send propagate signal to downstream

src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
OlegDokuka: fixed

src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
```diff
@@ -268,7 +269,8 @@ public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
                     if (requestsPending.get() > 0)
                         pollEvent.scheduleIfRequired();
                 })
-                .doOnCancel(() -> cancel(true));
+                .doOnTerminate(this::dispose)
+                .doOnCancel(this::dispose);
```
smaldini: can use doFinally

OlegDokuka: fixed

OlegDokuka:
@smaldini the last commit fixes the bug with racing between canceling and propagating of `onError` signal, so it may happen (and happened during Default Stage 😅) that the stream may hang infinitely because dispose of resources is executed faster than draining (`drainLoop`) for propagating an error in emitterProcessor Sink

Issue #72, opened 2018-12-13
`stopOnError` not observed on Producer sending


Sage-Pierce:
I have configured a KafkaSender to not stop on Errors (`stopOnError` == false) which is observed for Exceptions contained in callback Responses. However, this flag is _not_ observed for Exceptions resulting from the process of actually sending ProducerRecords:

https://github.com/reactor/reactor-kafka/blob/master/src/main/java/reactor/kafka/sender/internals/DefaultKafkaSender.java#L276

https://github.com/reactor/reactor-kafka/blob/master/src/main/java/reactor/kafka/sender/internals/DefaultKafkaSender.java#L283

https://github.com/reactor/reactor-kafka/blob/master/src/main/java/reactor/kafka/sender/internals/DefaultKafkaSender.java#L293

I have a case where I would like to continue processing Records on outbound Publishers even if a serialization Exception happens, which I can then handle downstream without the SendSubscriber fatally erroring out. This is useful when serialization is fallible; I use Avro which is dependent on network calls (but is recoverable) and there are cases where data is just not serializable, in which case I just want to log an Error and continue.

I believe this is a simple one-line change to have the `true` on line 293 swapped with `stopOnError`

bsideup:
Fixed in #73

PR #73, opened 2018-12-13
[#72] Apply stopOnError to Producer send Exceptions

Sage-Pierce:
A pass at implementing what is mentioned in #72


smaldini:
Simple and efficient, thanks a lot!

PR #75, opened 2019-01-17
Integrate BlockHound

bsideup:
This PR integrates [BlockHound](https://github.com/reactor/BlockHound) - blocking calls detector.

The build will fail because there are some blocking calls :)


build.gradle
```diff
@@ -101,6 +101,10 @@ configure(allprojects) { project ->
     jacoco {
       includes = ['reactor/kafka/sender/**', 'reactor/kafka/receiver/**']
     }
+
+    useJUnitPlatform {
```
ericbottard, line 105: Can't this use JUnit 5?

bsideup: not as part of this PR.

aneveu (approved):

PR #114, opened 2019-11-13
Improvements and updates of build

simonbasle:
- Bump gradle
 - use maven-publish + artifactory
 - prepare for a releaser script
 - cleanup build
 - set a `rootProject.name` in `settings.gradle`

This PR will need to be backported to `1.1.x` branch, while being careful that the move of `reactorCoreVersion` from `build.gradle` to `gradle.properties` is correctly backported with the right version inside.


LaoTsing:
in [DefaultKafkaSender.java](https://github.com/reactor/reactor-kafka/blob/8d14541bab6a8d41396bc2ef94ebc8e52a7f40e7/src/main/java/reactor/kafka/sender/internals/DefaultKafkaSender.java) please make **public** of `static class Response<T> implements SenderResult<T>` because not way to get this Response of `<T> Flux<SenderResult<T>> send(Publisher<? extends SenderRecord<K, V, T>> records);` in doOnEach function for looking both Complete and Error result

simonbasle:
@LaoTsing this is a PR that has nothing to do with any meaningful code change to `DefaultKafkaSender`, and is more of a cross-project effort. Please don't request random changes in unrelated PRs like that.

I don't see why `Response` should need to be public, as there is everything you need in the `SenderResult` interface.

LaoTsing:
I saw the documentation and examples of earlier versions, and in the doOnNext method the SenderResult actually returned, but now it doesn’t see it and the Result is returned, but down with error "Result class not accesable"

simonbasle:
Again, not the right place for such a question, open a separate more detailled issue and give more context + reproduction unit test.

PR #129, opened 2020-04-15
Replace `EmbeddedKafkaCluster` with Testcontainers

bsideup:


src/test/java/reactor/kafka/AbstractKafkaTest.java
```diff
@@ -194,33 +209,73 @@ protected void waitForTopic(String topic, int partitions, boolean resetMessages)
...
-        embeddedKafka.shutdownBroker(brokerId);
+    protected void assumeBrokerRestartSupport() {
+        Assume.assumeTrue("supports broker restart", false);
     }
 
-    protected void startKafkaBroker() {
-        embeddedKafka.startBroker(brokerId);
+    protected void shutdownKafkaBroker() {
+        assumeBrokerRestartSupport();
+        // TODO embeddedKafka.shutdownBroker(brokerId);
```
aneveu, line 251: not sure this TODO is still needed

src/test/java/reactor/kafka/AbstractKafkaTest.java
```diff
@@ -194,33 +209,73 @@ protected void waitForTopic(String topic, int partitions, boolean resetMessages)
...
     }
 
-    protected void restartKafkaBroker() {
-        embeddedKafka.restartBroker(brokerId);
-        waitForTopic(topic, partitions, false);
+    protected void startKafkaBroker() {
+        assumeBrokerRestartSupport();
+        // TODO embeddedKafka.restartBroker(brokerId);
```
aneveu, line 256: same here

src/test/java/reactor/kafka/AbstractKafkaTest.java
bsideup: Right. Replaced with `Assert.fail` so that (if we ever decide to add the broker restart scenario) it will remind us that we need to implement these methods 👍

.editorconfig
```diff
@@ -0,0 +1,22 @@
+# EditorConfig is awesome: http://EditorConfig.org
```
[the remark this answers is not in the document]
bsideup: I could, but had a bit of a trouble fighting with IDEA's desire to collapse all imports into `.*` :D

Ok, will extract the change...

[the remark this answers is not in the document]
bsideup: Extracted:
https://github.com/reactor/reactor-kafka/pull/134

PR #138, opened 2020-04-27
Cleanup code, remove `Scheduler` wrapper

bsideup:


src/main/java/reactor/kafka/receiver/internals/KafkaSchedulers.java
```diff
@@ -158,7 +61,7 @@ public final Thread newThread(Runnable runnable) {
             return t;
         }
 
-        static final class EmitterThread extends Thread implements NonBlocking {
+        static final class EmitterThread extends Thread {
```
simonbasle, line 64: is that thread really allowing blocking ? (ie can you confirm this change is on purpose, since the `single` thread by default would have produced nonblocking threads, but doesn't due to the explicit use of a factory)

bsideup: yes, this is an intentional change, because the emitter thread is doing a ton of blocking stuff (via Kafka's client), so it was incorrectly marked as non-blocking.

BlockHound barks on it as well :)

PR #139, opened 2020-04-28
Use `Scheduler` directly for the event loop

bsideup:


src/main/java/reactor/kafka/receiver/internals/ConsumerFlux.java
```diff
@@ -113,26 +102,67 @@
 
         pollEvent = new PollEvent();
         commitEvent = new CommitEvent();
-        scheduler = Schedulers.single(receiverOptions.schedulerSupplier().get());
+        eventScheduler = KafkaSchedulers.newEvent(receiverOptions.groupId());
```
bsideup, line 105: Note for the reviewers:

the diff is misleading. `scheduler` got moved to `DefaultKafkaReceiver`, and `eventScheduler` was moved to a field from here:
https://github.com/reactor/reactor-kafka/pull/139/files#diff-56dd574c6a295ffb097b9cfebce0390cL179

bsideup:
@aneveu FYI there is an interesting CI failure while running the benchmarks. Running them locally gives me *better* numbers than they were before (no surprise here, the event loop got dramatically simplified and has less overhead overall), so it may be some race condition or something.

I am battling it now, and may need to make some adjustments. If so, I will re-request your review 👍

bsideup:
Good news - I was able to pinpoint the failure and added a workaround (buffering of emitted `ConsumerRecords`)

Bad news - it seems that, previously, the issue was hidden behind `EmitterProcessor` and we need to fix the backpressure handling. This probably deserves a separate PR because it may not be trivial.

PR #141, opened 2020-04-30
Switch to Europium versions, -SNAPSHOT scheme

simonbasle:
See reactor/reactor#683


simonbasle:
depends on reactor/reactor-core#2140

PR #143, opened 2020-05-04
Dispose `ConsumerFlux` on cancel/error

bsideup:
This is a final step before we can encapsulate the logic inside the `Subscription`


src/main/java/reactor/kafka/receiver/internals/ConsumerFlux.java
```diff
@@ -269,6 +266,15 @@ public void dispose() {
         return Mono.empty();
     }
 
+    private void onError(Throwable t) {
+        CoreSubscriber<? super ConsumerRecords<K, V>> actual = this.actual;
+        try {
+            dispose();
+        } finally {
+            actual.onError(t);
```
simonbasle, line 274: suggestion: use `Exceptions.multiple` util to combine both errors.

PR #145, opened 2020-05-11
Introduce `ConsumerHandler` abstraction

bsideup:
Since `KafkaReceiver` needs access to the internal state and also
call methods on it, returning just `ConsumerFlux` isn't a viable
option.
This will allow us to asynchronously pre-initialize the state
and access it after receiving an instance of the handler.

### DIFF
--- a/src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
+++ b/src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
@@ -44,13 +44,14 @@ public class DefaultKafkaReceiver<K, V> implements KafkaReceiver<K, V> {
         this.receiverOptions = receiverOptions.toImmutable();
     }
 
-    private Mono<ConsumerHandler<K, V>> start() {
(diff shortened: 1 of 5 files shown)


src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java
```diff
@@ -25,135 +24,141 @@
...
+                return handler
+                    .receive()
+                    .flatMapIterable(it -> it)
+                    .map(record -> new ReceiverRecord<>(
+                        record,
+                        handler.toCommittableOffset(record)
+                    ))
+                    .doOnRequest(handler::handleRequest);
+            },
+            this::dispose
```
simonbasle, line 73: minor nitpick but I would just consider renaming the `dispose(ConsumerHandler)` method at the bottom to something along the lines of `disposeConsumerHandler` for clarity (when first reading the code my first thought was that there was a `Disposable` involved. granted, the ambiguity is not very strong...)

bsideup: 👍 changed to `cleanup`

PR #147, opened 2020-05-13
Rename `ConsumerFlux` -> `ConsumerEventLoop`

bsideup:


src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -53,80 +54,63 @@
...
-            Operators.error(actual, new IllegalStateException("Multiple subscribers are not supported for KafkaReceiver flux"));
-            return;
+        eventScheduler.schedule(new SubscribeEvent());
+
+        Duration commitInterval = receiverOptions.commitInterval();
+        if (!commitInterval.isZero()) {
+            switch (ackMode) {
+                case AUTO_ACK:
+                case MANUAL_ACK:
+                    eventScheduler.schedulePeriodically(
```
simonbasle, line 89: As a separate improvement I would consider grabbing the returned `Disposable` to allow for `stop()` method to dispose it, maybe?

bsideup: Isn't it handled by `Scheduler#dispose`?

simonbasle: I missed that in the `stop()` method, somehow. 👍

PR #151, opened 2020-05-19
Do not block when disposing the event loop

bsideup:


src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -132,69 +125,29 @@ private void onPartitionsRevoked(Collection<TopicPartition> partitions) {
...
-            for (int i = 0; i < maxRetries && !isConsumerClosed; i++) {
-                try {
-                    consumer.close();
-                    consumer = null;
-                    break;
-                } catch (Exception e) {
-                    if (i == maxRetries - 1) {
-                        log.warn("Consumer could not be closed", e);
-                    }
+                if (consumer == null) {
```
simonbasle, line 136: I wonder what the reason for a hard retry was, at that level... if it was actually relevant, then that capability is lost (due to the early switch of the `isActive` atomic), but as far as I can see there is some amount of retrying the kafka close involved in the `CloseEvent` too

garyrussell: @simonbasle @bsideup and I were looking at this code on Friday and I couldn't see any reason for it; in fact, now I look at the `KafkaConsumer` code, the close won't even do anything on subsequent attempts:

```java
    @Override
    public void close(Duration timeout) {
        if (timeout.toMillis() < 0)
            throw new IllegalArgumentException("The timeout cannot be negative.");
        acquire();
        try {
            if (!closed) {
                // need to close before setting the flag since the close function
                // itself may trigger rebalance callback that needs the consumer to be open still
                close(timeout.toMillis(), false);
            }
        } finally {
            closed = true;
            release();
        }
    }
```

The `closed` boolean is set in the `finally` block.

PR #154, opened 2020-06-02
Use `Schedulers#immediate` as a default scheduler

bsideup:


src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -238,26 +238,30 @@ public void run() {
             try {
                 if (commitArgs != null) {
                     if (!commitArgs.offsets().isEmpty()) {
-                        inProgress.incrementAndGet();
                         switch (ackMode) {
                             case ATMOST_ONCE:
                                 consumer.commitSync(commitArgs.offsets());
```
bsideup, line 243: Synchronous, no need to increment/decrement I assume

src/test/java/reactor/kafka/receiver/internals/MockReceiverTest.java
```diff
@@ -596,15 +596,17 @@ public void manualAckCommitBatchSize() {
...
-            long offset = r.receiverOffset().offset();
-            if (offset < 10) {
-                r.receiverOffset().acknowledge();
-                if (((offset + 1) % batchSize) == 0)
-                    lastCommitted.set(offset);
-            } else
-                uncommittedMessages.add(r);
-            verifyCommit(r, lastCommitted.get());
-            return Mono.just(r);
+            return Mono.fromCallable(() -> {
```
bsideup, line 599: I spent ~2 days debugging the tests after applying that small change.

"never block the emitter thread" makes a lot more sense now 😂

src/test/java/reactor/kafka/sender/internals/MockTransactionTest.java
```diff
@@ -133,10 +133,10 @@ public void transactionalReceiveAndSend() throws Exception {
...
 
-        assertEquals(transactionCount, producer.beginCount);
-        assertEquals(transactionCount, producer.commitCount);
-        assertEquals(0, producer.abortCount);
-        assertEquals(transactionCount, producer.sendOffsetsCount);
+        assertThat(producer.beginCount).as("beginCount").isGreaterThanOrEqualTo(transactionCount);
+        assertThat(producer.commitCount).as("commitCount").isGreaterThanOrEqualTo(transactionCount);
+        assertThat(producer.abortCount).as("abortCount").isZero();
+        assertThat(producer.sendOffsetsCount).as("sendOffsetsCount").isEqualTo(transactionCount);
     }
```
garyrussell, line 140: @bsideup This should fix the flakiness - yes, it's possible you will get less then 10 per poll.

Unfortunately, there is no `poll.min.records`, you can only approximate it with `fetch.min.bytes`.

bsideup: Thanks! This is the assumption I had but wasn't sure about, appreciate the clarification 👍

PR #156, opened 2020-06-22
Cleanup deprecations and warnings

bsideup:


src/main/java/reactor/kafka/sender/internals/DefaultKafkaSender.java
```diff
@@ -196,7 +197,7 @@ public void subscribe(CoreSubscriber<? super Object> s) {
                 .then();
     }
 
-    private <T> Flux<SenderResult<T>> transaction(Publisher<? extends SenderRecord<K, V, T>> transactionRecords, UnicastProcessor<Object> transactionBoundary) {
+    private <T> Flux<SenderResult<T>> transaction(Publisher<? extends SenderRecord<K, V, T>> transactionRecords, FluxIdentityProcessor<Object> transactionBoundary) {
```
simonbasle, line 200: is the `transactionBoundary` really useful? the `concatWith(Mono.create`) a few line below is also very strange

bsideup: yes, I was also WTFing at `concatMap(Mono.create(`, but didn't want to touch/change the current logic.

FTR "Sender" refactoring is planned for M2 (while "Receiver" was a target for M1), where I will cleanup it anyways

PR #158, opened 2020-07-01
Use `immediate` Sender scheduler by default

bsideup:


bsideup:
@simonbasle after checking why the tests were failing I figured that `Producer#send` could deadlock everything (exactly what happened on Travis). To fix that, I had to move `send` to a separate `Scheduler` (which also makes BlockHound happy, btw :))

src/test/java/reactor/kafka/sender/internals/MockSenderTest.java
```diff
@@ -421,7 +420,6 @@ public void sendWithResponseStopOnError() {
                     .recordWith(() -> sendResponses)
                     .expectError(InvalidTopicException.class)
                     .verify(Duration.ofMillis(DEFAULT_TEST_TIMEOUT));
-        assertEquals(maxInflight, outgoing.onNextCount.get());
```
bsideup, line 424: the assertion wasn't correct (breaks due to prefetching). We have `producer.enableInFlightCheck();` a few lines earlier that will ensure that `maxInflight` is correct.

src/test/java/reactor/kafka/sender/internals/MockSenderTest.java
```diff
@@ -150,7 +150,6 @@ public void sendNoResponseFailure() {
         StepVerifier.create(sender.createOutbound().send(outgoing.producerRecords()).then())
                     .expectError(InvalidTopicException.class)
                     .verify(Duration.ofMillis(DEFAULT_TEST_TIMEOUT));
-        assertEquals(maxInflight, outgoing.onNextCount.get());
```
bsideup, line 153: the assertion wasn't correct (breaks due to prefetching). We have `producer.enableInFlightCheck();` a few lines earlier that will ensure that `maxInflight` is correct.

PR #164, opened 2020-08-10
[fixup] Overlooked Sinks API migration for 1.3.0-M2

bsideup:


src/main/java/reactor/kafka/receiver/internals/ConsumerHandler.java
```diff
@@ -63,7 +62,8 @@
 
     private final ConsumerEventLoop<K, V> consumerEventLoop;
 
-    private final FluxIdentityProcessor<ConsumerRecords<K, V>> processor = Processors.more().multicast(1);
+    @SuppressWarnings("deprecation")
+    private final EmitterProcessor<ConsumerRecords<K, V>> processor = EmitterProcessor.create(1);
```
bsideup, line 66: I need `EmitterProcessor` here because later we pass `FluxSink` to `ConsumerEventLoop` that uses `FluxSink#onRequest` / `FluxSink#requestedFromDownstream()`

PR #179, opened 2020-10-08
Sync with reactor-core (`EmitFailureHandler`)

bsideup:


src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -164,6 +164,11 @@ private void onPartitionsRevoked(Collection<TopicPartition> partitions) {
             });
     }
 
+    @Override
+    public boolean onEmitFailure(SignalType signalType, Sinks.Emission emission) {
+        return isActive.get();
```
simonbasle, line 169: the parkNanos was removed, could be conditionally added back while also checking the `signalType`? this effectively can loop indefinitely in case of `TERMINATED` for instance, which would be wasteful

simonbasle (changes requested):
this PR should wait a bit for the last round of breaking change in the `Emission` class https://github.com/reactor/reactor-core/pull/2419

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
bsideup: Nice catch!

aneveu (approved):

simonbasle (approved):

simonbasle:
@bsideup `Emission` renamed to `EmitResult` has been merged, snapshots incoming

simonbasle (approved):

simonbasle (approved):

PR #184, opened 2020-11-30
Fix ImmutableSenderOptions.closeTimeout()

garyrussell:
- argument passed in was ignored


bsideup:
@garyrussell it looks like the issue was introduced long time before 1.3.x. Could you please use 1.2.x as a baseline for the fix?

garyrussell:
@bsideup Done.

bsideup:
@garyrussell thank you! 👍

PR #186, opened 2020-12-03
Switch to GitHub Actions for CI

bsideup:
This includes removing `.travis.yml` file


ericbottard (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
+name: CI
+
+on:
+    pull_request: {}
+    push: { branches: [ master, "1.2.x" ] }
```
simonbasle, line 5: is the second one a regex pattern? don't we want to also build `1.1.x` branch, and in the future build `1.3.x` branch without having to touch this file ?

simonbasle (changes requested):

.github/workflows/ci.yml
bsideup: no, it is a concrete value ("1.2.x"). GHA support patterns, I can change to to a pattern if you wish :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
+name: CI
+
+on:
+    pull_request: {}
+    push:
+      branches:
+        - master
+        - 1.*
```
simonbasle, line 8:
```suggestion
        - *.*.x
```

ericbottard: Beware if you start the pattern with a star: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

bsideup: I would not do that, yeah. 2.0 is going to be a big change anyways, so it is okay to add a line to the CI file by the time it is happening :)

simonbasle: how about this safer pattern:
```suggestion
        - [0-9]+.[0-9]+.x
```

simonbasle: Last suggestion (what I'm doing in core's latest PR, turns out the quotes are necessary here too):
```suggestion
        - "[0-9].[0-9]+.x"
```

simonbasle (approved):

simonbasle (changes requested):
there's a travis.yml file that needs to be removed.

simonbasle (approved):

Issue #183, opened 2020-11-09
Out of memory on DefaultKafkaReceiver with high batch size, returned by poll


ivyazmitinov:
During testing on the prod with the amount of data in Kafka about 250Gb and configs resulting in 50-100K records in one poll batch we faced the OOM on 2GB heap in about 30 minutes of usage. After an investigation we found out that the root cause in the default `prefetch` parameters of the `publishOn` and default queues in `ConsumerHandler#sink`. 

## Expected Behavior

With the following configs:
```
max.poll.records: 100000
fetch.max.wait.ms: 1000
max.partition.fetch.bytes: 5242880 # 50Mb
fetch.max.bytes: 104857600   # 100Mb
```
And the following usage:
```
KafkaReceiver.create(receiverOptions)
                   .receiveExactlyOnce(transactionManager)
                   .concatMap(recordsBatch -> actualizeBatchTransactional(recordsBatch, transactionManager), 1)
```
It is expected to only one batch at the time to be requested from the `KafkaReceiver` due to the `concatMap` semantics.

## Actual Behavior

256 batches are requested asynchronously due to implementation of  `DefaultKafkaReceiver`. It is caused by usage of `publishOn` with the default `prefetch` parameter in all `receive*` methods, which is propagated to the `ConsumerEventLoop#onRequest` instead of `concatMap`'s 1, effectively causing it to make 256 polls.
Also, as turned out, default size of `ConsumerHandler#sink` also causes memory issues, probably becase incoming `ConsumerRecords` are double buffered and can't be effectively garbage collected: in the underlying `UnicastProcessor` queue and the `publishOn` queue

## Steps to Reproduce

It is tricky to reproduce locally, but, possibly, could be done with mocking of the `kafka.clients.consumer.Consumer`

## Possible Solution

We fixed it by passing the prefetch value to the underlying `publishOn` and reducing the `ConsumerHandler#sink`'s queue size to 1. See the [following](https://github.com/ivyazmitinov/reactor-kafka/commit/447e4ff69a1bd4287d6d1a93ebdc603ffd98a748) commit.
Also, on the `1.2.4.RELEASE` only `publishOn` fix was required, `EmitterProcessor` worked fine.

## Your Environment

* Reactor version(s) used: 3.4.0
* Other relevant libraries versions (eg. `netty`, ...): io.projectreactor.kafka:reactor-kafka:1.3.0
* JVM version (`javar -version`): openjdk version "1.8.0_265"
[...]

PR #199, opened 2021-03-03
Fix Partition Resumes

### LINKED ISSUE #190: Need help on backpressure and auto pause / resume during high load for reactive consumer
Hi All,

I have a reactive kafka consumer which consumes data from a topic and pushes it to an endpoint. However, these endpoints are flaky and are error prone. I have circuit breakers which trip on failures and I induce an artificial delay `Mono.delay` to slow down the pipeline. By theory, reactive backpressure should kick in and pause the kafka consumers right? I see that If i set my max.poll.interval.ms to low number, it does not pause the consumer and it results in rebalancing. How can I make sure that the Kafka consumer applies back-pressure and pauses the consumption until the downstream messages are done processing?
Here's my consumer code:

```
protected Disposable pollRecordsWithAutoCommit(KafkaReceiver<byte[], T> receiver) {
    return receiver
        .receiveAutoAck()
        .onErrorStop()
        .flatMap(
            consumerRecordFlux -> processFluxStream(consumerRecordFlux),
            jobConfiguration.getConsumerParallelPower())
        .onErrorContinue(
            // ALL uncaught exceptions will be published here.
            // TODO:: post to DLQ with retry count or possibly other meta data
            // `o` is the KAFKA record
            (throwable, o) -> {
              log.error(
                  "Error while processing record. [error:{}, record:{}]", throwable.toString(), o);
            })
        .subscribe();
```

processFluxStream has `Mono.delay` code to slow down incase of circuit breaker errors.

How do I go about linking the reactive pipeline to pause and resume on these signals and also avoid rebalancing issues?
I read about an earlier issue which had the same problem but that does not have a solution.

Thanks.

### LINKED ISSUE #198: reactor.core.scheduler.Schedulers KafkaScheduler worker in group main failed with an uncaught exception
I get a memory leak when my consumers work on dev enviroments, without any load

## Expected Behavior
memory not leak

## Actual Behavior
memory not leak

## Steps to Reproduce

1) create receiver
```java
  @Bean
  KafkaReceiver<String, UserProfile> upsertUserProfile() {
    var topic = kafkaTopics.getUpsertUserProfile();
    var servers = kafkaProperties.getKafkaBootstrapAddress();
    var properties = createReceiverProperties(servers, UserProfile.class);

    log.info("createKafkaReceiver: topic={}, consumerGroupName={}, properties={}",
             topic,
             servers,
             properties
    );

    ReceiverOptions<String, UserProfile> options = ReceiverOptions.create(
      Map.of(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, servers,
        ConsumerConfig.GROUP_ID_CONFIG, kafkaProperties.getGroupId(),
        ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest",
        ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class,
        ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, JsonDeserializer.class,
        JsonDeserializer.CLASS_NAME, UserProfile.class,
        ConsumerConfig.FETCH_MAX_BYTES_CONFIG, 26214400,
        ConsumerConfig.MAX_PARTITION_FETCH_BYTES_CONFIG, 524288
      )
    );

    return KafkaReceiver.create(
      options
        .commitBatchSize(300)
        .subscription(Collections.singleton(topic))
    );
  }
```
2) read messages
```java
@Service
@AllArgsConstructor
public class UserProfileConsumer {

  private final KafkaReceiver<String, UserProfile> upsertUserProfile;
  private final UserProfileService userProfileService;


  @PostConstruct
  private void init() {

    userProfileConsumer();
  }

  private void userProfileConsumer() {
    upsertUserProfile
      .receiveAutoAck()
      .concatMap(r -> r)
      .flatMap(record -> {
        var message = record.value();
        return userProfileService.save(message);
      })
      .onErrorContinue((e, item) -> log.error("userProfileConsumer: error" + item, e))
      .retryWhen(
        RetrySpec.indefinitely()
      )

      .subscribe();
  }


}
```

memory leak 24hours

and jstack 

[jstack.txt](https://github.com/reactor/reactor-kafka/files/6029655/stack.txt)

## Your Environment

* Reactor version(s) used: io.projectreactor.kafka:reactor-kafka:1.3.1
* Other relevant libraries versions (eg. `netty`, ...): io.projectreactor:reactor-bom:2020.0.3
* JVM version (`java -version`): 14
* Java env: xmx 1gb
* OS and version (eg `uname -a`): Linux bdbac3b775aa 5.4.0-54-generic #60-Ubuntu SMP Fri Nov 6 10:37:59 UTC 2020 x86_64 Linux

garyrussell:
`ConsumerEventLoop.PollEvent` should not resume partitions that it, itself,
has not paused.

Prior to https://github.com/reactor/reactor-kafka/commit/d908b815bf5a8be41b840a272e6766ad8754783d
pause/resume due to backpressure was protected by a boolean but it still unconditionally
resumed all partitions.

- reintroduce the boolean and only resume if we paused
- keep track of user-paused partitions and exclude them from resuming

Also fixes #190
Also fixes #198

### DIFF
--- a/reactor-kafka-tools/src/test/resources/log4j.properties
+++ b/reactor-kafka-tools/src/test/resources/log4j.properties
@@ -1,4 +1,4 @@
-# Copyright (c) 2016-2018 Pivotal Software Inc, All Rights Reserved.
+# Copyright (c) 2020 VMware Inc, All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/reactor-kafka-tools/src/test/resources/log4j.properties
+++ b/reactor-kafka-tools/src/test/resources/log4j.properties
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 VMware Inc, All Rights Reserved.
+# Copyright (c) 2016-2018 Pivotal Software Inc, All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
+++ b/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
@@ -119,7 +119,7 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
 
     void onRequest(long toAdd) {
         if (log.isDebugEnabled()) {
-            log.debug("onRequest.toAdd:" + toAdd);
+            log.debug("onRequest.toAdd: " + toAdd);
         }
         Operators.addCap(REQUESTED, this, toAdd);
         pollEvent.schedule();
@@ -251,13 +251,13 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
                             if (!pausedByUs.getAndSet(true)) {
                                 this.pausedByUser.addAll(consumer.paused());
                                 consumer.pause(consumer.assignment());
-                                log.debug("Paused 1");
+                                log.debug("Paused - awaiting transaction");
                             }
                         }
                     } else if (!pausedByUs.getAndSet(true)) {
                         this.pausedByUser.addAll(consumer.paused());
                         consumer.pause(consumer.assignment());
-                        log.debug("Paused 2");
+                        log.debug("Paused - back pressure");
                     }
 
                     ConsumerRecords<K, V> records = consumer.poll(pollTimeout);
--- a/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
+++ b/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
@@ -118,6 +118,9 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
     }
 
     void onRequest(long toAdd) {
+        if (log.isDebugEnabled()) {
+            log.debug("onRequest.toAdd:" + toAdd);
+        }
         Operators.addCap(REQUESTED, this, toAdd);
         pollEvent.schedule();
     }
@@ -237,27 +240,27 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
                             if (pausedByUs.getAndSet(false)) {
                                 Set<TopicPartition> toResume = new HashSet<>(consumer.assignment());
                                 toResume.removeAll(this.pausedByUser);
+                                this.pausedByUser.clear();
                                 consumer.resume(toResume);
+                                log.debug("Resumed");
                             }
                         } else {
                             if (!pausedByUs.getAndSet(true)) {
-                                this.pausedByUser.clear();
                                 this.pausedByUser.addAll(consumer.paused());
                                 consumer.pause(consumer.assignment());
+                                log.debug("Paused 1");
                             }
                             schedule();
                         }
                     } else if (!pausedByUs.getAndSet(true)) {
-                        this.pausedByUser.clear();
                         this.pausedByUser.addAll(consumer.paused());
                         consumer.pause(consumer.assignment());
+                        log.debug("Paused 2");
                     }
 
                     ConsumerRecords<K, V> records = consumer.poll(pollTimeout);
                     if (isActive.get()) {
-                        if (r > 1 || commitEvent.inProgress.get() > 0) {
-                            schedule();
-                        }
+                        schedule();
                     }
 
                     Operators.produced(REQUESTED, ConsumerEventLoop.this, 1);
--- a/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
+++ b/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
@@ -119,7 +119,7 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
 
     void onRequest(long toAdd) {
         if (log.isDebugEnabled()) {
-            log.debug("onRequest.toAdd: " + toAdd);
+            log.debug("onRequest.toAdd {}", toAdd);
         }
         Operators.addCap(REQUESTED, this, toAdd);
         pollEvent.schedule();
@@ -265,8 +265,11 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
                         schedule();
                     }
 
-                    Operators.produced(REQUESTED, ConsumerEventLoop.this, 1);
-                    sink.emitNext(records, ConsumerEventLoop.this);
+                    if (!records.isEmpty()) {
+                        Operators.produced(REQUESTED, ConsumerEventLoop.this, 1);
+                        log.debug("Emitting {} records, requested now {}", records.count(), r);
+                        sink.emitNext(records, ConsumerEventLoop.this);
+                    }
                 }
             } catch (Exception e) {
                 if (isActive.get()) {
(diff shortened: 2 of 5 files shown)


garyrussell:
cc/ @artembilan - I couldn't tag you for a review - maybe you aren't in this org?

artembilan:
True. Here are people of this org: https://github.com/orgs/reactor/people. 
But that still does not prevent us from reviewing. 
Although looks like the merge is still only their prerogative 😄

artembilan (approved):

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -230,12 +236,22 @@ public void run() {
                     long r = requested;
                     if (r > 0) {
                         if (!awaitingTransaction.get()) {
-                            consumer.resume(consumer.assignment());
+                            if (pausedByUs.getAndSet(false)) {
+                                Set<TopicPartition> toResume = new HashSet<>(consumer.assignment());
+                                toResume.removeAll(this.pausedByUser);
+                                consumer.resume(toResume);
```
simonbasle, line 242: is it relevant to keep the `pausedByUser` collection full after this point?
ie. all the branches where `pausedByUs == false` will clear the collection and refill it, but that clear is thus done in a lazy fashion

src/test/java/reactor/kafka/receiver/KafkaReceiverTest.java
```diff
@@ -1190,6 +1190,29 @@ public void abortTransaction() throws Exception {
         checkConsumedMessages(count, count * 3);
     }
 
+    @Test
+    public void userPause() throws InterruptedException {
```
simonbasle, line 1194: does the test really cover externally paused topic partitions? ie. would there be a way to have 2 partitions paused and verify that the `PollEvent` distinguishes between the two more explicitly?

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -230,12 +236,22 @@ public void run() {
                     long r = requested;
                     if (r > 0) {
                         if (!awaitingTransaction.get()) {
-                            consumer.resume(consumer.assignment());
+                            if (pausedByUs.getAndSet(false)) {
```
simonbasle, line 239: can two `run()` execute concurrently? ie could it be that while we atomically switch `pausedByUs` to `false` here, the line 245 below gets executed? or is it all serialized by the `awaitingTransaction` and `requested` somehow ?

simonbasle (commented):
need a few clarifications

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
garyrussell: Good point; yes, it can be cleared here.

src/test/java/reactor/kafka/receiver/KafkaReceiverTest.java
garyrussell: Yes; more test coverage is needed; I originally wrote this test just to demonstrate that the paused partition was incorrectly resumed. 

The notion of user Vs. our pauses only came to me while fixing the bug, but I didn't enhance the test to cover that use case.

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
garyrussell: I don't believe so, the scheduler is a `Schedulers.newSingle()`.

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -116,6 +118,9 @@
     }
 
     void onRequest(long toAdd) {
+        if (log.isDebugEnabled()) {
+            log.debug("onRequest.toAdd:" + toAdd);
```
artembilan, line 122: Probably need a space after colon to make it more readable in the logs

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -218,30 +223,46 @@ public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
...
+                                this.pausedByUser.addAll(consumer.paused());
+                                consumer.pause(consumer.assignment());
+                                log.debug("Paused 1");
+                            }
                         }
-                    } else {
+                    } else if (!pausedByUs.getAndSet(true)) {
+                        this.pausedByUser.addAll(consumer.paused());
                         consumer.pause(consumer.assignment());
+                        log.debug("Paused 2");
```
artembilan, line 260: What the info does it bring for end-users in the logs?
I mean this one and the previous `Paused 1` . 
Doesn't look like very useful...

src/test/java/reactor/kafka/AbstractKafkaTest.java
```diff
@@ -61,11 +47,24 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
 public abstract class AbstractKafkaTest {
 
     public static final int DEFAULT_TEST_TIMEOUT = 60_000;
 
-    private static final KafkaContainer KAFKA = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:5.5.1"))
+    private static final KafkaContainer KAFKA = new KafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:6.1.0"))
```
artembilan, line 67: Can't we use a latest image here?
I mean is it possible to have it without version at all, which means `latest`?

reactor-kafka-tools/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2016-2018 Pivotal Software Inc, All Rights Reserved.
```
artembilan, line 1: It's kinda weird to have a Copyright in the non-Java files, but OK if it is what we have so far in this project.
Anyway: I see you have modified Copyright to VMware in other place. So, this one might be affected, too.

On the other hand the Log4J 1.x is out of life for a while already: https://logging.apache.org/log4j/1.2/ 😄 

Different issue though...

artembilan (changes requested):

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -218,30 +223,46 @@ public void onPartitionsRevoked(Collection<TopicPartition> partitions) {
...
 
                     ConsumerRecords<K, V> records = consumer.poll(pollTimeout);
                     if (isActive.get()) {
-                        if (r > 1 || commitEvent.inProgress.get() > 0) {
-                            schedule();
-                        }
+                        schedule();
                     }
 
                     Operators.produced(REQUESTED, ConsumerEventLoop.this, 1);
```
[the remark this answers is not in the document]
garyrussell: I had that before, but restored it because `onRequest()` was never called (it's called downstream from the sink) so we never resumed the consumer.

That might have been on an earlier iteration of this patch; I'll look again tomorrow.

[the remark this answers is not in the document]
garyrussell: See https://github.com/reactor/reactor-kafka/issues/183#issuecomment-796229889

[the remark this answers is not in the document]
garyrussell: After further review, I had to pull that change because `KafkaReceiverTests.userPause()` now fails, because now there is no thread to call `onRequest()` to resume the consumer. Need to give this some more thought.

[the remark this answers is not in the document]
garyrussell: I was mistaken; that code looks fine; there is something else going on that I need to track down tomorrow.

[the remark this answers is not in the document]
peterfigure: Thanks

PR #209, opened 2021-03-31
Remove JCenter from build

simonbasle:
﻿See reactor/reactor#695.


violetagg (approved):

simonbasle:
~(NB: this will probably need some adaptation in 1.3.x)~

looks like the forward-merge is trivial 👍

PR #210, opened 2021-04-14
GitHub Actions Release Process: signing, artifactory, sonatype

simonbasle:
This commit changes the release process to be based on GitHub Actions.

CI is already performed on GHA, but that workflow will now only run
for pull_request trigger.

We define helper tasks and methods in the releaser.gradle file to help
interpret the version number. The trigger is pushing a commit to an
active maintenance branch.

The new workflow uses GitHub Environments to manage access to secrets,
especially in the case of milestones and releases as it introduces
signing of artifacts in these cases. It is split into 4 jobs, 3 of
which are conditional and depend on a GitHub Environment.

Notably:
- Add a function to qualify a version number as one of RELEASE,
MILESTONE, SNAPSHOT or BAD
- Add a GitHub Actions oriented task, `qualifyVersionGha`, which
produces gha step output for the current project version. `versionType`
is an application of the above function and `fullVersion` is the plain
project's `version`
- tag releases and milestones using `git` commands in their respective
jobs
- Activate the JFrog Artifactory plugin only if a password is defined
(including in subprojects)
- Add the `signing` plugin, activated only if a signing key is defined,
requiring that signing be performed if a publish task is invoked other
than mavenLocal (can be forced with `-PforceSigning`)
- Add a Maven publication repository if Sonatype OSS credentials are
defined, plus a local `mock` repository for debugging of build
- cleaned up the developers list in the pom.xml

See reactor/reactor#694.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,12 +1,6 @@
 name: CI
-
 on:
     pull_request: {}
-    push:
-      branches:
-        - master
-        - "[0-9].[0-9]+.x"
-
 jobs:
   build:
     name: build
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,109 @@
+name: publish
+on:
+  push:
+    branches: # For branches, better to list them explicitly than regexp include
+      - master
+      - 1.2.x
+jobs:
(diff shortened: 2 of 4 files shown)


violetagg (approved):

simonbasle:
- [ ] prepare environments and secrets
 - [ ] disable build on bamboo
 - [ ] forward-merge to 1.3.4

PR #217, opened 2021-04-20
Add Trace Diagnostics for Consumers

garyrussell:
Log current positions and committed offsets on assignment.

Tested with spring-kafka.


simonbasle (approved):
LGTM. I'm assuming the `position` and `committed` methods are fine to run from within the listener (kafka consumer nested calls)...

garyrussell:
Yes, this is called on the event loop thread when the partitions are assigned.

PR #222, opened 2021-06-29
[build] Asciidoctor-convert 1.5.8.1 not found, bump to 1.5.12

simonbasle:


simonbasle:
Encountered this while working on another PR: the project wasn't even building, complaining that plugin's version `1.5.8.1` couldn't be found in configured repositories...
Apparently, it works with `1.5.12`. I've verified that the `asciidoctor` task still works with the upgrade and generates html5 and pdf reference guides correctly ✅

OlegDokuka (approved):
LGTM

PR #223, opened 2021-06-29
[chores] Copyright switch Pivotal to VMware, polish headers

simonbasle:
- copyright header to VMware instead of Pivotal
 - update the pom content (emails, org)
 - in copyright headers, have date range harmonized on 2016-2021
 - polish copyright header: url indentation, blank line after license

See reactor/reactor#682.


simonbasle:
⚠️ do not merge, I will need to rollback changes to the start of the year range to keep original year when possible (ie start year can vary).

OlegDokuka (approved):

simonbasle:
#226 is a follow-up PR that fixes the headers for files introduced in 1.3.x

Issue #224, opened 2021-06-30
The ReceiverPartition's passed to the listener has seek methods, but not one for seekToTimestamp


deepakemani-maersk:
The current Reactor kafka in the consumer has support to seek to an offset, seek to Beginning and seek to end but **The Reactor Kafka does not have support in the consumer to seek to a timestamp**

API reference : [https://projectreactor.io/docs/kafka/release/api/reactor/kafka/receiver/ReceiverPartition.html](url)

## Desired solution

It will be good to have this feature say a method called seekToTimeStamp (long timeStamp) in ReceiverPartition class.
This would be used like below when we create the KafkaReciever using the **addAssignListener**() method
```java
long timeStamp = 1624989510813L;
receiverOptions = receiverOptions
            .addAssignListener(partitions -> partitions.forEach(p -> p.seekToTimeStamp (timeStamp))) 
            .subscription(Collections.singleton(topic));
KafkaReceiver.create(receiverOptions).receive().subscribe();
```

## Considered alternatives

At the moment, there are no alternatives and only option that is working for us is to seek for an offset.

garyrussell:
This is not so easy, since `consumer.offsetsForTimes()` (needed to seek to offsets representing a specific time) is a blocking operation and we can't run a blocking operation on the consumer thread, but we must do the seeks before the listener exits.

I need to give this some thought as to how this functionality could be made available.

garyrussell:
Given that the consumer API is not thread-safe, we have no choice but to perform the blocking operation on the consumer event loop thread. This will work, but BlockHound will complain if you use this technique. 

I think we will just have to live with that, and add an exclusion via a `BlockHoundIntegration` to prevent BlockHound from complaining.

PR #225, opened 2021-06-30
GH-224: Support seekToTimestamp

garyrussell:
Resolves https://github.com/reactor/reactor-kafka/issues/224

### DIFF
--- a/src/main/java/reactor/kafka/receiver/ReceiverPartition.java
+++ b/src/main/java/reactor/kafka/receiver/ReceiverPartition.java
@@ -51,6 +51,8 @@ public interface ReceiverPartition {
 
     /**
      * Seek to the topic partition offset that is greater than or equal to the timestamp.
+     * If there are no matching records, {@link #seekToEnd()} is performed. See
+     * {@link org.apache.kafka.clients.consumer.Consumer#offsetsForTimes(java.util.Map)}.
      * @param timestamp the timestamp.
      * @since 1.3.5
      */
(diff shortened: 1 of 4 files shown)


src/test/java/reactor/kafka/receiver/KafkaReceiverTest.java
```diff
@@ -192,6 +192,39 @@ public void seekToOffset() throws Exception {
...
+        Flux<? extends ConsumerRecord<Integer, String>> kafkaFlux = receiver
+                .receive()
+                .doOnError(e -> log.error("KafkaFlux exception", e));
+        CountDownLatch latch = new CountDownLatch(count);
+        subscribe(kafkaFlux, latch);
+
+        waitForMessages(latch);
+        this.receivedRecords.forEach(list -> list.forEach(rec ->
+                assertThat(rec.timestamp()).isGreaterThanOrEqualTo(t1)));
+        assertThat(this.receivedRecords.stream().flatMap(list -> list.stream()).count()).isEqualTo(count);
```
simonbasle, line 224: nit: add `.as("sum of list sizes")` before .isEqualTo as a way to see what the int represents in case of failure (otherwise the generic assertion error is just "expected x, got y")

src/main/java/reactor/kafka/receiver/ReceiverPartition.java
```diff
@@ -48,6 +48,13 @@
      */
     void seek(long offset);
 
+    /**
+     * Seek to the topic partition offset that is greater than or equal to the timestamp.
```
simonbasle, line 52: should it clearly specify that if no such offset is found, this is equivalent to `seekToEnd()`?

simonbasle (changes requested):
looks good to me. a couple nitpicks.
NB: the PR needs a resync with base branch, feel free to rebase and force-push then address the comments (or do a merge, that should work too ;) )

simonbasle (approved):

Issue #215, opened 2021-04-16
Using sendTransactionally makes blockhound agent throws BlockingOperationError


guillaumelamirand:
Hi,

I am using reactor kafka to read and publish to kafka and when I have added reactor.blockhound agent to check any blocking call, it appears that using ` kafkaSender.sendTransactionally` is detected as doing blocking call. 

## Expected Behavior
I am thinking that the blocking call is required because of the transaction however blockhound shouldn't complain about it.

## Actual Behavior

Exception is thrown:
```
Operator called default onErrorDropped
reactor.core.Exceptions$ErrorCallbackNotImplemented: reactor.blockhound.BlockingOperationError: Blocking call! jdk.internal.misc.Unsafe#park
Caused by: reactor.blockhound.BlockingOperationError: Blocking call! jdk.internal.misc.Unsafe#park
	at java.base/jdk.internal.misc.Unsafe.park(Unsafe.java)
	at java.base/java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:234)
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(AbstractQueuedSynchronizer.java:1079)
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1369)
	at java.base/java.util.concurrent.CountDownLatch.await(CountDownLatch.java:278)
	at org.apache.kafka.clients.producer.internals.TransactionalRequestResult.await(TransactionalRequestResult.java:69)
	at org.apache.kafka.clients.producer.KafkaProducer.initTransactions(KafkaProducer.java:599)
	at reactor.kafka.sender.internals.DefaultKafkaSender.lambda$new$0(DefaultKafkaSender.java:102)
	at reactor.core.publisher.MonoCallable.subscribe(MonoCallable.java:56)
	at reactor.core.publisher.InternalMonoOperator.subscribe(InternalMonoOperator.java:64)
	at reactor.core.publisher.MonoCacheTime.subscribeOrReturn(MonoCacheTime.java:143)
	at reactor.core.publisher.Mono.subscribe(Mono.java:4084)
	at reactor.core.publisher.FluxConcatArray$ConcatArraySubscriber.onComplete(FluxConcatArray.java:208)
	at reactor.core.publisher.FluxConcatArray.subscribe(FluxConcatArray.java:80)
[...]

PR #231, opened 2021-07-13
GH-215: Remove Invalid Test

garyrussell:
GH-215 added this test, but it is invalid because `stopOnError(false)` makes no
sense for a transactional send, since the transaction will be rolled back anyway.
If this occurs before the record is actually sent, no record is available at the
consumer (which is not transactional anyway so it can "see" uncommitted records
(which is why it worked most of the time).

Remove the test; the same PR added verification that the blocking calls were made
on the correct thread.

Also add failed test artifact upload for PR builds.


simonbasle (approved):
sorry, review fell through the cracks, approved

PR #235, opened 2021-07-29
GH-233: Wake the consumer when the back pressure is relieved

### LINKED ISSUE #233: When fetching from partitions is paused, consumer.poll(pollTimeout) method should not be called
In [ConsumerEventLoop#PollEvent](https://github.com/reactor/reactor-kafka/blob/553aa5c92a25431a8fecce647d78b8e94b32cca5/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java#L254) there are two options when fetching from the requesting partitions can be paused:

- [awaiting for transaction](https://github.com/reactor/reactor-kafka/blob/553aa5c92a25431a8fecce647d78b8e94b32cca5/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java#L274)
- [back pressure](https://github.com/reactor/reactor-kafka/blob/553aa5c92a25431a8fecce647d78b8e94b32cca5/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java#L280)

In both cases after `consumer.pause(consumer.assignment())`, `consumer.poll(pollTimeout)`method is called which will run until timeout is reached and return no records because fetching from the partitions is paused. Such behavior slows down receiver performance.

Example:
When poll timeout is 10 seconds and record request came immediately after back pressure pause, we had to wait 10 seconds before run method started again and fetching from partitions was resumed.
```
2021-07-28 17:26:15.011  DEBUG 3484 --- [-19f4ff15730e-1] r.k.r.internals.ConsumerEventLoop        : Started run method
2021-07-28 17:26:15.011  DEBUG 3484 --- [-19f4ff15730e-1] r.k.r.internals.ConsumerEventLoop        : Requested: 0
2021-07-28 17:26:15.011 DEBUG 3484 --- [-19f4ff15730e-1] r.k.r.internals.ConsumerEventLoop        : Paused - back pressure
2021-07-28 17:26:15.046 DEBUG 3484 --- [amController-10] r.k.r.internals.ConsumerEventLoop        : onRequest.toAdd 1
2021-07-28 17:26:25.011  DEBUG 3484 --- [-19f4ff15730e-1] r.k.r.internals.ConsumerEventLoop        : Finished run method
2021-07-28 17:26:25.012  DEBUG 3484 --- [-19f4ff15730e-1] r.k.r.internals.ConsumerEventLoop        : Started run method
2021-07-28 17:26:25.012  DEBUG 3484 --- [-19f4ff15730e-1] r.k.r.internals.ConsumerEventLoop        : Requested: 1
2021-07-28 17:26:25.012 DEBUG 3484 --- [-19f4ff15730e-1] r.k.r.internals.ConsumerEventLoop        : Resumed
....
```

## Expected Behavior
When fetching from partitions is paused, `consumer.poll(pollTimeout)` method should not be called.

* Reactor version(s) used: 1.3.4

arina-ielchiieva:
Resolves #233

### DIFF
--- a/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
+++ b/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
@@ -137,6 +137,9 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
             log.debug("onRequest.toAdd {}", toAdd);
         }
         Operators.addCap(REQUESTED, this, toAdd);
+        if (pollEvent.isPaused()) {
+            consumer.wakeup();
+        }
         pollEvent.schedule();
     }
 
@@ -281,7 +284,13 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
                         log.debug("Paused - back pressure");
                     }
 
-                    ConsumerRecords<K, V> records = consumer.poll(pollTimeout);
+                    ConsumerRecords<K, V> records;
+                    try {
+                        records = consumer.poll(pollTimeout);
+                    } catch (WakeupException e) {
+                        records = ConsumerRecords.empty();
+                    }
+
                     if (isActive.get()) {
                         schedule();
                     }
@@ -304,6 +313,10 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
             if (!this.scheduled.getAndSet(true)) {
                 eventScheduler.schedule(this);
             }
+        }
+
+        boolean isPaused() {
+            return pausedByUs.get();
         }
     }
 
--- a/src/test/java/reactor/kafka/receiver/KafkaReceiverTest.java
+++ b/src/test/java/reactor/kafka/receiver/KafkaReceiverTest.java
@@ -1278,6 +1278,35 @@ public class KafkaReceiverTest extends AbstractKafkaTest {
         flux.dispose();
     }
 
+    @Test
+    public void wakeUpReceiverWhenBackPressureIsRelieved() throws Exception {
+        int count = 10;
+        receiverOptions = receiverOptions
+            .pollTimeout(Duration.ofMinutes(1))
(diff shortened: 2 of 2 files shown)


src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -266,22 +270,31 @@ public void run() {
                                 toResume.removeAll(this.pausedByUser);
                                 this.pausedByUser.clear();
                                 consumer.resume(toResume);
+                                isPaused.set(false);
```
simonbasle, line 273: at each point `isPaused` is set to a boolean, `pausedByUs` has already been set to that same boolean. I'm not saying I'm sure this is a suitable replacement, but wondering if there's a good reason not to use `pausedByUs`? The only thing I can think of is that `pausedByUs` is set to `true` in more cases than the ones for which we want to `consumer.wakeup()` above at line 142.

simonbasle (changes requested):
@arina-ielchiieva would you be able to demonstrate the issue (and thus the fix) in a unit test?

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
garyrussell: Yes; we can use `pausedByUs` by pushing it up to the enclosing class.

However, I don't believe we can consider this PR until the CLA is "signed".

arina-ielchiieva: Good point, made changes to use `pausedByUs` as indicator to wake up the consumer.
Signed CLA as well.

arina-ielchiieva:
@simonbasle  / @garyrussell thanks for the review. Applied required changes and added unit test.

simonbasle (approved):
LGTM. what do you think @garyrussell?

garyrussell (approved):

Issue #259, opened 2022-01-07
Consumer gets stuck after a rebalance when using out of order commits


pdanelson:
A consumer rebalance, eg. a consumer leaving a multi consumer group, may trigger the remaining consumers to have an ever increasing deferred commit count which eventually leads to an indefinitely paused consumer if using the out of order commits feature.

## Expected Behavior

Consumer should continue consuming events and committing even after a rebalance.

## Actual Behavior

After a rebalance, the deferred commit count starts increasing until it reaches the max deferred commits limit at which point the consumer pauses and never recovers.

## Steps to Reproduce

```java
@Test
void reproCase() {
    var consumerProps = Map.<String, Object>of(
        ConsumerConfig.GROUP_ID_CONFIG, TEST_TOPIC,
        ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, getKafkaServer(),
        ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest"
    );
    var consumerOpts = ReceiverOptions.<String, String>create(consumerProps)
        .commitInterval(Duration.ofSeconds(2L))
        .maxDeferredCommits(1000) // Just to demonstrate that consumer pauses and never recovers, the deferred count itself would grow unbounded
        .withKeyDeserializer(new StringDeserializer())
        .withValueDeserializer(new StringDeserializer())
        .subscription(Collections.singleton(TEST_TOPIC)); // Test topic should have 2+ partitions

    var subscription1 = KafkaReceiver.create(consumerOpts)
        .receive()
        .subscribe();
    
    var subscription2 = KafkaReceiver.create(consumerOpts)
        .receive()
        .flatMap(record -> Mono.delay(Duration.ofMillis(100L * RandomUtils.nextInt(1, 5))) // Simulate async event processing
                .doOnSuccess(__ -> record.receiverOffset().acknowledge()),
            10)
        .subscribe();
    
    Thread.sleep(20_000L); // Deferred count stays consistently in the single digits during this period
    subscription1.dispose(); // Force a re-balance
    
    Thread.sleep(100_000L); // Deferred count starts to grow until it maxes out the deferred commits limit during this period
    subscription2.dispose();
}
```

## Possible Solution

[...]

PR #260, opened 2022-01-11
Fix Out of Order Commits After Rebalance

garyrussell:
Resolves https://github.com/reactor/reactor-kafka/issues/259

Consumer could be permanently paused if a partition is reassigned and
duplicate records received.

- purge deferred and uncommitted offset commits when a partition is revoked
- protect against duplicate entries in both maps

Also fix a small race where `isPaused()` returns false in `onRequest()` but the
consumer is about to be paused, preventing wake up - only an issue with long
poll timeouts.


rancherz:
Looks good

Issue #206, opened 2021-03-19
Add Micrometer Integration (KafkaClientMetrics)


garyrussell:
See https://stackoverflow.com/questions/66706766/spring-boot-kafka-kafka-metrics-not-available-in-actuator-prometheus/66711272#66711272

Zane-XY:
Hi @garyrussell 
can we prioritize this?

garyrussell:
Sure; contributions are always welcome.

Does the solution in the stack overflow answer not work for you?

Zane-XY:
@garyrussell 
Yeah, I tried it the solution in stackoverflow. The process and commit happens in the `doOnNext` operator. In my case, I want the poll and process happens in the asynchronous manner, so I use `flatMap` and subscribed on a different scheduler than the poll thread.  But I got an error saying something like `You must call one of receive*() methods before using doOnConsumer`.

garyrussell:
?

The consumer is not created until you subscribe to the flux created by the `receive()` (regardless of the upstream topology); once you have done so, you can register the metrics.

ageorgousakis:
I have tried the solution on the stack overflow. It is working but when there is problem with the poll flow and a commit is failed the metrics stop collecting information. It seems Kafka creates another consumer and the doOnConsumer method is not called with the new consumer

garyrussell:
Registering the meters is not part of the pipeline; you would need to add code to detect the error, unregister the old metrics and register the new ones.

I will take a look at integrating this soon, if nobody from the community offers to do it.

ageorgousakis:
Can you give us a sample code snippet  how to register the meters in reactor kaka without being stopped in case of any errors. With `spring-kafka` there is a solution can we have the something similar in `reactor-kafka`?

garyrussell:
Something like this:

```java
KafkaReceiver<String, String> receiver = KafkaReceiver.create(ro);
AtomicReference<Consumer<String, String>> consumerToUnregister = new AtomicReference<>();
receiver.receive()
	.doOnNext(rec -> {
		System.out.println(rec.value());
		throw new RuntimeException("test");
	})
	.doOnSubscribe(subs -> {
		receiver.doOnConsumer(consumer -> {
			consumerListener.consumerAdded("myConsumer", consumer);
			consumerToUnregister.set(consumer);
			return Mono.empty();
		}).subscribe();
	})
	.onErrorResume(e -> {
[...]

PR #262, opened 2022-02-08
Add Micrometer Support for Consumer and Producer

garyrussell:
Resolves https://github.com/reactor/reactor-kafka/issues/206#issuecomment-1006119883

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -149,6 +149,7 @@ configure(allprojects) { project ->
     api libs.kafka
     api libs.reactor.core
     optionalApi libs.micrometer.core
+// 2.0.x   optionalApi libs.micrometer.binders
 
     compileOnly libs.jsr305
     testCompileOnly libs.jsr305
--- a/build.gradle
+++ b/build.gradle
@@ -111,6 +111,12 @@ configure(allprojects) { project ->
 
   [compileJava, compileTestJava]*.options*.compilerArgs = ['-Xlint:all,-options']
 
+  java {
+    registerFeature('optional') {
+        usingSourceSet(sourceSets.main)
+    }
+  }
+
   test {
     maxHeapSize = "1024m"
     testLogging {
@@ -142,6 +148,7 @@ configure(allprojects) { project ->
   dependencies {
     api libs.kafka
     api libs.reactor.core
+    optionalApi libs.micrometer.core
 
     compileOnly libs.jsr305
     testCompileOnly libs.jsr305
--- a/src/main/java/reactor/kafka/receiver/ImmutableReceiverOptions.java
+++ b/src/main/java/reactor/kafka/receiver/ImmutableReceiverOptions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/reactor/kafka/receiver/ImmutableReceiverOptions.java
+++ b/src/main/java/reactor/kafka/receiver/ImmutableReceiverOptions.java
@@ -16,6 +16,14 @@
 
 package reactor.kafka.receiver;
 
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.serialization.Deserializer;
+import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
+import reactor.util.annotation.Nullable;
+
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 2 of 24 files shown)


build.gradle
```diff
@@ -127,8 +128,10 @@ configure(allprojects) { project ->
 
   dependencies {
     compileOnly "com.google.code.findbugs:jsr305:$googleJsr305Version"
+    compileOnly "io.micrometer:micrometer-core:$micrometerVersion"
```
artembilan, line 131: Not `optional`?

src/main/java/reactor/kafka/receiver/ImmutableReceiverOptions.java
```diff
@@ -30,13 +38,6 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
-import org.apache.kafka.clients.consumer.ConsumerConfig;
-import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.common.config.ConfigException;
-import org.apache.kafka.common.serialization.Deserializer;
-import reactor.core.scheduler.Scheduler;
-import reactor.core.scheduler.Schedulers;
```
artembilan, line 38: Is it correct for this project to rearrange imports?

Not related, but in Spring Integration 6.0 I'm going to follow IDE settings from Spring Framework: https://github.com/spring-projects/spring-integration/issues/3686
From where I believe the import order is exactly like it was before your refactoring here.

artembilan (changes requested):

build.gradle
garyrussell: This version of gradle doesn't seem to support it; we really need to upgrade to a recent version but my gradle/groovy skills are not up to it.

src/main/java/reactor/kafka/receiver/ImmutableReceiverOptions.java
garyrussell: We had nothing like that when this syntax used to work 

    compile ("com.fasterxml.jackson.core:jackson-core:$jacksonVersion", optional)

https://github.com/spring-projects/spring-kafka/blob/5c54ad02339ab1849f98fd3e785f811c6deb17c5/build.gradle#L198

artembilan: Yes. I don't mind, but turns out Gradle has decided to go further and above limited number of Maven dependency variants.
But unfortunately they liquidated an `optional` as out-of-the-box at all...

artembilan (approved):
LGTM.

Still don't have permissions to merge PRs in this project 😄

build.gradle
garyrussell: Yes but a lot has changed since then `compile` etc have been deprecated - there are other changes too - e.g. maven publishing.

simonbasle: @garyrussell @artembilan upgrading this project to Gradle 7.4 is on my todo list. made a bit difficult by the fact that it lagged on 5.x indeed...

src/main/java/reactor/kafka/receiver/ImmutableReceiverOptions.java
simonbasle: in core we do use `compileOnly` (but indeed these don't appear in the pom, see https://github.com/reactor/reactor-core/issues/2262)

garyrussell:
Need to rework this on Monday due to gradle upgrade (thanks @simonbasle ).

artembilan (approved):

simonbasle (commented):
nit: added `maxDeferredCommits` to equals and hashCode looks like a separate issue.

I was also going to bring up that with upcoming Micrometer 2.0, maybe there's little sense in adding Micrometer 1 metrics so late in the cycle. But that's not going to be GA until the end of the year, and given the metrics code in this change I don't think the migration to Micrometer 2 will be terribly hard. so 👍

garyrussell:
Another thought would be to just add the listener interface and leave off the micrometer implementations out and just add the code to the reference manual since it's so small; the user can copy/paste.

WDYT?

migroskub:
> Another thought would be to just add the listener interface and leave off the micrometer implementations out and just add the code to the reference manual since it's so small; the user can copy/paste.
> 
> WDYT?

I think it misses the whole point. As a user, I'd like to just have an easy way of making this work. If you expect most of the users to copy-paste the metrics, I think it worth it to let them be there, and the more advanced needs will implement the api.

simonbasle:
@garyrussell yeah that could be an intermediate way to improve things in 1.3.x.

Let me list a few factoids and constraints:
 - Micrometer 2.x is around the corner, GA late 2022
 - there are no metrics at all in reactor-kafka 1.3.x right now
 - reactor-kafka 1.3.x cannot use Micrometer 2.x (we need to maintain compatibility with reactor-core 3.4.x which uses  Micrometer 1)
 - there could be a reactor-kafka 1.4.0, part of the 2022.0 upcoming release cycle (or compatible with it) that uses Micrometer 2.x (late 2022)

@migroskub I understand your perspective, but as stated above there are _no_ metrics right now. if you're adamant about metrics _right now_ then you could be considered an advanced user and would get an interface to listen and implement metrics with some effort, albeit minimal.
Then in 1.4.0 all users will get something more future-proof working off the shelf.

Just playing devil's advocate here. I don't want to rush implementing metrics with Micrometer 1.x when Micrometer 2.x is around the corner.

PR #267, opened 2022-02-18
[chores] Update the build / Gradle 7.4 / GHA polish

simonbasle:
- Polish ci and workflow: action hardening, prepare for Gradle 7 bump
- Switch to Gradle 7.4
- Remove jacoco aggregation custom code
- Switch to version catalog
- update libraries specific to this project
- Remove seemingly useless metrics lib


.github/workflows/ci.yml
```diff
@@ -1,25 +1,24 @@
...
+        distribution: 'temurin'
         java-version: 8
-    - uses: eskatos/gradle-command-action@v1
+    - uses: gradle/gradle-build-action@937999e9cc2425eddc7fd62d1053baf041147db7 # renovate: tag=v2
       name: gradle
       with:
         arguments: build javadoc
-        wrapper-cache-enabled: true
-        dependencies-cache-enabled: true
-        configuration-cache-enabled: true
```
artembilan, line 19: Why do you remove all of these caches?
They are applied only for the current PR, so subsequent commits take less time to rebuild the PR.

artembilan (changes requested):
GH Action shows some JavaDocs problems, but I guess it is not relevant to this change.
We can address later on.

Thank

artembilan (approved):
OK. Thank you for all the explanations.
LGTM

Issue #127, opened 2020-04-08
Commit retry has no backoff


CodeMyWorld:
When kafka client commits failed, client will retry immediately after the failed commit request. And after 100 times retry within 10 ms, the comsumer flux is terminated, the consumer will not poll new messages anymore
## Expected Behavior

Retry should have backoff, instead of retrying immediately
## Actual Behavior

Retry commit immediately. By checking the log, the consumer thread retried to commit 100 times within 10ms. This time is not enough for Kafka client failover in a clustered environment.
## Steps to Reproduce

Start the consumer, and then stop one of kafka broker.

## Possible Solution

Add some backoff before next try here:
https://github.com/reactor/reactor-kafka/blob/2e96f8e3a621c7675d33af63015624b81830fbb0/src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java#L587
## Your Environment

* Reactor version(s) used:  3.2.0
* Other relevant libraries versions (eg. `netty`, ...):
* JVM version (`javar -version`): 11
* OS and version (eg `uname -a`): Ubuntu 18.04

garyrussell:
Please explain further; it's not clear to me why the commits are retried at all in this code.

If a rebalance is in progress, it means we have probably lost the partitions anyway (and they may have been assigned to a different instance) so, IMHO, it's completely wrong to retry offset commits.

garyrussell:
```
    /**
     * Returns the maximum number of consecutive non-fatal commit failures that are tolerated.
     * For manual commits, failure in commit after the configured number of attempts fails
     * the commit operation. For auto commits, the receive Flux is terminated.
     * @return maximum number of commit attempts
     */
    @NonNull
    int maxCommitAttempts();
```

Commits are retried only for `RetriableCommitFailedException`.

I very much doubt that a failure to commit due to a rebalance would be considered retriable; by definition, we would have lost the partitions and they may have been reassigned to another instance.

billydh:
@CodeMyWorld how did you figure out that the time period in which those 100 commit attempts are tried is 10ms? I tried to looking in `ReceiverOptions`, `MutableReceiverOptions` but could not find that 10ms time period.

billydh:
@garyrussell is there a way to test the `maxCommitAttempts` at all? i.e. is it possible to write a unit test in my application to simulate that `RetriableCommitFailedException`? my unit tests are all using `EmbeddedKafka`.

CodeMyWorld:
[...]

PR #268, opened 2022-03-01
Add Delay for Commit Retries

garyrussell:
Resolves https://github.com/reactor/reactor-kafka/issues/127

Previously, the commit task scheduled itself immediately for retries.

- Add `ReceiverProperty.commitRetryInterval` (default 500ms).
- Schedule the commit task according to that interval
- Pause the consumer while retrying
- Make `KafkaReceiverTest.testAutoCommitFailureScenarios` more robust
- Fix `ChaosConsumerFactory` to not commit requested, for reliable redelivery
- Add a new Mock test for the functionality (also already covered by real and
  `MockReceiver` tests)
- Fix a flaky `ConsumerEventLoopTest`
- Fix an NPE in some `MockConsumer` tests

### DIFF
--- a/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
+++ b/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2020-2022 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -274,7 +274,7 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
                     long r = requested;
                     boolean pauseForDeferred = this.maxDeferredCommits > 0
                         && this.commitBatch.deferredCount() >= this.maxDeferredCommits;
-                    if (pauseForDeferred) {
+                    if (pauseForDeferred || commitEvent.retrying) {
                         r = 0;
                     }
                     if (r > 0) {
@@ -298,6 +298,8 @@ class ConsumerEventLoop<K, V> implements Sinks.EmitFailureHandler {
                         consumer.pause(consumer.assignment());
                         if (pauseForDeferred) {
                             log.debug("Paused - too many deferred commits");
+                        } else if (commitEvent.retrying) {
+                            log.debug("Paused - commits are retrying");
                         } else {
                             log.debug("Paused - back pressure");
                         }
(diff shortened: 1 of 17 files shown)


simonbasle:
I need to actually review this, but I noticed that some modified files haven't had their license header updated, and some have. I think I should activate `spotless` to check license headers in all projects including reactor-kafka, but that's a separate thing

garyrussell:
@simonbasle Sorry about that; I thought I caught them all, but obviously not. A tool would be nice; we have a custom gradle task in some of our projects.

simonbasle:
@garyrussell I have opened a separate PR that activates the Spotless plugin with license header checks. It can be merged before that one and we should then get a simple way to fix all these headers (`./gradlew spotlessApply`)

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -360,6 +362,7 @@ boolean isPaused() {
         private final AtomicBoolean isPending = new AtomicBoolean();
         private final AtomicInteger inProgress = new AtomicInteger();
         private final AtomicInteger consecutiveCommitFailures = new AtomicInteger();
+        private boolean retrying;
```
simonbasle, line 365: we check `this.retrying`, which is a simple `boolean`, in places where we also check `isPending` (`AtomicBoolean`). Should `retrying` also be an `AtomicBoolean` ?

src/test/java/reactor/kafka/receiver/internals/CommitRetryTests.java
```diff
@@ -0,0 +1,128 @@
...
+        assertTrue(subscribeLatch.await(10, TimeUnit.SECONDS));
+        assertTrue(receiveLatch.await(10, TimeUnit.SECONDS));
+        received.get(0).receiverOffset().commit();
+        assertTrue(commitLatch.await(10, TimeUnit.SECONDS));
+        verify(consumer, times(5)).commitAsync(any(), any());
+        Map<TopicPartition, OffsetAndMetadata> commits = new HashMap<>();
+        commits.put(new TopicPartition("foo", 0), new OffsetAndMetadata(1L));
+        verify(consumer, times(5)).commitAsync(eq(commits), any());
+        disposable.dispose();
+    }
```
simonbasle, line 126: would it be possible to add a test that validates the configured delay is used? maybe integrate that in the test above, since for now it only configures `commitInterval` and `maxCommitAttempts`?

simonbasle (changes requested):

src/test/java/reactor/kafka/receiver/internals/CommitRetryTests.java
garyrussell: Good catch; thanks!

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
garyrussell: Good catch; thanks!

simonbasle (approved):
side note: `./gradlew spotlessApply` should generally only touch files that have been changed by the PR (or when checked out locally, files that differ between the local branch and the `origin/main` branch)

garyrussell:
@simonbasle Thanks; that explains my spotless issues; I generally don't keep my origin/main up to date; my local `main` branch tracks `upstream/main` so I can easily fetch and pull changes from the primary repo.

I will back-out those bogus copyright changes.

simonbasle (approved):

simonbasle (approved):

PR #273, opened 2022-03-10
Prettify readme by removing redundant path

apffa:


simonbasle (approved):
that makes sense, thanks

Issue #266, opened 2022-02-17
Rebalancing during offset commit results in event processing termination


wjur:
## Expected Behavior
When the rebalancing happens during the commit, I expect a temporary, retryable error and a commit retry.
If the retry is not possible, then I expect the uncommitted messages to be read again at some point of time
(maybe by a different consumer).

## Actual Behavior
In our production environment, it happens that some service instances stop reading events from Kafka after the rebalancing.
In the logs we observe a couple warn and error messages. Most notably:
```text
Member [censored] sending LeaveGroup request to coordinator 
[censored] (id: [censored] rack: null) due to the consumer is being closed
```
```text
Commit failed
```
```text
reactor.core.Exceptions$ErrorCallbackNotImplemented: 
    org.apache.kafka.common.errors.RebalanceInProgressException: 
    Offset commit cannot be completed since the consumer is undergoing 
    a rebalance for auto partition assignment. You can try completing 
    the rebalance by calling poll() and then retry the operation.
  Caused by: org.apache.kafka.common.errors.RebalanceInProgressException: 
    Offset commit cannot be completed since the consumer is undergoing
     a rebalance for auto partition assignment. You can try completing
      the rebalance by calling poll() and then retry the operation.
```
```text
org.apache.kafka.common.metrics.Metrics - 
  Closing reporter org.apache.kafka.common.metrics.JmxReporter
org.apache.kafka.common.metrics.Metrics - 
  Metrics reporters closed
```

## Possible root cause
It seems that when `ConsumerCoordinator::sendOffsetCommitRequest` throws `RebalanceInProgressException` ([source](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L1206)) it is not handled in any way (no retry, no swallowing), but propagated and as a result closes the `sink` that is used to push polled events.

## The code path that misbehaves
I went through the code and think I know what is happening here. The first log, especially the part 
`the consumer is being closed` led me to `ConsumerCoordinator::close`. I checked where it is called,
and it turns out the only interesting place is the `DefaultKafkaReceiver::withHandler` method ([source](https://github.com/reactor/reactor-kafka/blob/6cf11d9c4386bbeaec59e68ebee9b7c47b12c5c1/src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java#L128))

[...]

Issue #240, opened 2021-09-10
Concurrent Processing with Partition-Based Ordering w/ Manual Acknowledgement - Race Condition on Rebalance


damianlarson:
Maybe this is just an issue with our specific implementation of this, but we're seeing unexpected behavior that's causing problems when rebalancing happens.

In our current process, we're using concurrent processing with partition-based ordering with the intention being to consume each partition sequentially on the same set of threads, as described in the reactor-kafka docs. In our process, we check to see if we're assigned the partition before processing--if so, we continue, if not, we return--and we also check after processing if we still have the partition assigned. If we have the partition assigned after processing, we make the manual acknowledgement, and reactor will make the commits when it needs to. What we're seeing is that when partitions are revoked, the Flux continues to have messages. The problem here lies in the asynchronicity, if a partition is re-assigned a partition while the Flux is still processing a message, we will process it successfully, see that we own the partition, and acknowledge the offset, which will acknowledge all of the offsets before, even though we didn't actually process those messages. Sample code is shown below, but it appears to be a race condition where the Flux is not re-initialized (or purged) when Kafka takes the partitions, even though they should be.

## Expected Behavior

Flux is drained/reinitialized when Kafka revokes partitions. Docs for ReceiverOptions indicate that `when group management is used, revoke listeners are invoked before every rebalance operation`, which if true, the above behavior is not in line with Kafka's docs for the `ConsumerRebalanceListener`, which indicate that `in eager rebalancing, it will always be called at the start of a rebalance and after the consumer stops fetching data.`

## Actual Behavior

Messages persist on the Flux and are successfully processed, which will cause a race condition if the same partition is later re-assigned.

## Steps to Reproduce

Our consumer is as follows:
```java
ReceiverOptions<Integer, String> options = ReceiverOptions.create(props);
receiverOptions = options.subscription(getTopicAndReplica(topic))
        .addAssignListener(partitions -> {
            assignedPartitions = partitions;
            log.info("{} Partitions Assigned {}", name, AOUtilities.printPartitions(partitions));
        })
        .addRevokeListener(partitions -> {
            assignedPartitions.clear();
            log.info("{} Partitions Revoked {}", name, AOUtilities.printPartitions(partitions));
        })
        .commitInterval(Duration.ofSeconds(1))
        .commitBatchSize(1)
        .closeTimeout(Duration.ofSeconds(30))
        .pollTimeout(Duration.ofSeconds(5));
        
[...]

PR #281, opened 2022-06-14
GH-266: Delay Rebalance Until Pipeline Drained

garyrussell:
Resolves https://github.com/reactor/reactor-kafka/issues/266
Resolves https://github.com/reactor/reactor-kafka/issues/240

Due to the async nature of reactive workloads, group rebalancing
can occur while records remain from the previous poll(s).
This causes issues with commit failures and record duplication.

Add a configurable maxiimum delay which will be applied when a rebalance
occurs, to wait for records in the pipeline to be processed and let
the rebalance occur at that time.

During the delay, periodically commit any available offsets with the
interval for these commits also being confgurable.


garyrussell:
Also tested with this (and debug logging):

```java
@SpringBootApplication
public class Rkgh266Application {

	private static final Logger log = LoggerFactory.getLogger(Rkgh266Application.class);

	private final CountDownLatch latch = new CountDownLatch(1);

	public static void main(String[] args) {
		SpringApplication.run(Rkgh266Application.class, args);
	}

	@Bean
	public ApplicationRunner runner1() {
		return args -> {
			SenderOptions<String, String> so = SenderOptions.create(
					Map.of(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092",
							ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class,
							ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class,
							ProducerConfig.ACKS_CONFIG, "all",
							ProducerConfig.TRANSACTIONAL_ID_CONFIG, "tx-"));
			KafkaSender<String, String> sender = KafkaSender.create(so);
			Flux<SenderRecord<String, String, String>> recs = Flux.just(pr("foo"), pr("bar"), pr("fail"));
			Flux<Flux<SenderResult<String>>> flux = sender.sendTransactionally(Flux.just(recs));
			flux.subscribe(res -> res.subscribe(result -> {
				System.out.println(result.recordMetadata());
			}));
			Thread.sleep(5000);
		};
	}

	private SenderRecord<String, String, String> pr(String value) {
		return SenderRecord.create("rk266", 0, null, null, value, value + ".corr");
	}

	@Bean
	public ApplicationRunner runner2() {
		return args -> {
			receiver();
			this.latch.await(60, TimeUnit.SECONDS);
			receiver();
		};
	}

	private void receiver() {
		ReceiverOptions<String, String> ro = ReceiverOptions.<String, String> create(
				Map.of(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092",
						ConsumerConfig.GROUP_ID_CONFIG, "rk266",
						ConsumerConfig.MAX_POLL_RECORDS_CONFIG, 3,
						ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest"))
				.withKeyDeserializer(new StringDeserializer())
				.withValueDeserializer(new StringDeserializer())
				.addAssignListener(assignments -> log.info("Assigned: " + assignments))
				.commitBatchSize(1)
				.addAssignListener(parts -> this.latch.countDown())
				.subscription(Collections.singletonList("rk266"));
		KafkaReceiver.create(ro)
				.receive()
				.publishOn(Schedulers.single(), 1)
				.doOnNext(record -> {
					System.out.println(record.value() + "@" + record.offset());
					record.receiverOffset().acknowledge();
					try {
						Thread.sleep(3000);
					}
					catch (InterruptedException e) {
						Thread.currentThread().interrupt();
					}
				})
				.subscribe();
	}

	@Bean
	public NewTopic topic() {
		return TopicBuilder.name("rk266").partitions(2).replicas(1).build();
	}

}
```

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
```diff
@@ -149,8 +149,29 @@ private void onPartitionsRevoked(Collection<TopicPartition> partitions) {
...
+                long maxDelayRebalance = receiverOptions.maxDelayRebalance().toMillis();
+                if (isActive.get() && maxDelayRebalance > 0) {
+                    long interval = receiverOptions.commitIntervalDuringDelay();
+                    int inPipeline = commitEvent.commitBatch.getInPipeline();
+                    if (inPipeline > 0 || this.awaitingTransaction.get()) {
+                        long end = maxDelayRebalance + System.currentTimeMillis();
+                        do {
+                            try {
+                                log.debug("Rebalancing; waiting for {} records in pipeline", inPipeline);
+                                Thread.sleep(interval);
```
simonbasle, line 163: would it be better to schedule so task `interval` milliseconds in the future, rather than sleep/blocking ? or this is guaranteed to be executed on a `Thread` that can be blocked?

reactor-kafka-samples/src/test/java/reactor/kafka/samples/SampleScenariosTest.java
```diff
@@ -96,8 +97,9 @@ public void kafkaSinkChain() throws Exception {
         sinkChain.runScenario();
         waitForMessages(expected, received1);
         List<Person> expected2 = new ArrayList<>();
-        for (Person p : expected)
+        for (Person p : expected) {
```
simonbasle, line 100: nitpick: there seems to be unrelated formatting changes in this PR

simonbasle (changes requested):

src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java
bensilvan: From what I know, rebalance happens on the consumption thread, and during rebalance, consumer is not polling messages. I think its better to block the thread while the thread pool continues to process and acknowledge messages. This way we prevent redundant thread pool tasks transfers*

garyrussell: We must block it to prevent Kafka from completing the rebalance while the pipeline is drained.

I believe it is a blockable thread because `commitSync()` (which is optionally called) blocks.

reactor-kafka-samples/src/test/java/reactor/kafka/samples/SampleScenariosTest.java
simonbasle: as far as I can see, no meaningful change but formatting in that file (and marking a field above as `final`), so I would lean towards removing the formatting just in this file

simonbasle: nevermind that, there's a method call added to a method call chain below in `kafkaTransform`. I don't mind terribly, just a habit of avoiding large formatting changes in unmodified lines in other reactor projects.

simonbasle (approved):

PR #300, opened 2022-10-20
Improve ack throughput with deferred commits

### LINKED ISSUE #299: Improve acknowledgement throughput when utilising deferred commits under high load
## Motivation

Currently acknowledgement time for a record grows linearly with the sizes of the uncommitted and deferred offsets queues due to them being implemented as linked lists which need to be checked for the presence of the offset for each acknowledgement. This can become a throughput bottleneck for high load workloads, especially since acknowledgement is a [synchronized method](https://github.com/reactor/reactor-kafka/blob/86d545613fbcb005587693a8b310d3581cb62fd7/src/main/java/reactor/kafka/receiver/internals/CommittableBatch.java#L48) and thus can not be invoked concurrently from multiple threads.

For example, based on a quick benchmark on my own machine, a deferred commit queue size of 100 000 records would already cap the throughput of the consumer to a maximum 3 000 records per second as the contains check of the linked list alone would take ~0.3 milliseconds for each acknowledgement.

Screenshot 2022-10-19 at 21 31 40

## Desired solution

Based on the logic that the deferred and uncommitted offsets queues already get cleared [when partitions are revoked](https://github.com/reactor/reactor-kafka/blob/86d545613fbcb005587693a8b310d3581cb62fd7/src/main/java/reactor/kafka/receiver/internals/CommittableBatch.java#L94), there should(?) not be any possibility of duplicate offsets and it should be possible to remove the expensive checks for element presence and make the acknowledgement an effectively constant time operation.

**NB!** If the above assumption about duplicates is wrong then I would expect that there might be a bug present in the feature regarding draining the pipeline before rebalancing as the records in pipeline counter is incremented for every polled record (no duplicate offset check), but decremented only if the acknowledgement is not a duplicate.

## Considered alternatives

In case there is another edge case where duplicate offsets may occur and the presence checks can not be removed, replacing the linked list based implementation with a tree set based implementation would at least make the acknowledgments scale logarithmically not linearly with the queue sizes (while also maintaining offsets in order and not requiring an explicit sort when [committing](https://github.com/reactor/reactor-kafka/blob/86d545613fbcb005587693a8b310d3581cb62fd7/src/main/java/reactor/kafka/receiver/internals/CommittableBatch.java#L118)) the offsets.

## Additional context

Our workload is a highly concurrent one where we split records from each partition into hundreds of parallel rails, each of which can be temporarily delayed due to transient network issues while the rest of the rails continue consumption, thus temporarily creating high numbers of deferred commits. Those transient issues tend to end up being distributed uniformly across the parallel rails and we are not overly concerned about duplicate consumption, so we prefer to keep a high deferred commits ceiling to avoid consumer pauses in favour of higher throughput. Currently, this can become problematic though, since high numbers of deferred commits will make us hit the throughput bottleneck that is hard to recover from.

pdanelson:
Replace LinkedList with TreeSet for deferred and uncommitted offset queues to ensure acknowledgements scale at O(log n) as opposed to O(n) in respect to queue sizes when using deferred commits.

Fixes #299.

I also benchmarked 3 potential implementations to compare the performance (microseconds per scenario, lower is better) by simulating the acknowledgement and committing of 100, 1 000, 10 000 and 100 000 records in random order while using deferred commits ([benchmark code](https://gist.github.com/pdanelson/3deaa0f2d460e0ff698b94c3dda6a4ad)):
- LINKED_LIST - The existing implementation
- TREE_SET - The PR implementation
- NO_CHECK - Implementation without any checks for duplicate offsets in deferred or uncommitted queues (simply the removal of the `.contains(...)` invocations in `CommitableBatch.updateOffset` and `CommitableBatch.addUncommitted` from the existing implementation)

Screenshot 2022-10-20 at 12 59 52

The results show the tree set implementation having a significant performance gain over the existing implementation at all tested counts (especially so at the higher end with a 500x difference in performance) and a small, but noticeable performance loss over the implementation without any duplicate checks at all. However, since I am not 100% sure of there being no possible race conditions where duplicates may appear and the performance difference is somewhat negligible, I opted to go for the tree set implementation which keeps the behaviour identical.

### DIFF
--- a/src/main/java/reactor/kafka/receiver/internals/CommittableBatch.java
+++ b/src/main/java/reactor/kafka/receiver/internals/CommittableBatch.java
@@ -25,20 +25,20 @@ import reactor.core.publisher.MonoSink;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.SortedSet;
+import java.util.TreeSet;
 
 class CommittableBatch {
 
     private static final Logger log = LoggerFactory.getLogger(CommittableBatch.class);
 
     final Map<TopicPartition, Long> consumedOffsets = new HashMap<>();
-    final Map<TopicPartition, List<Long>> uncommitted = new HashMap<>();
-    final Map<TopicPartition, List<Long>> deferred = new HashMap<>();
+    final Map<TopicPartition, SortedSet<Long>> uncommitted = new HashMap<>();
+    final Map<TopicPartition, SortedSet<Long>> deferred = new HashMap<>();
     private final Map<TopicPartition, Long> latestOffsets = new HashMap<>();
     boolean outOfOrderCommits;
     private int batchSize;
@@ -47,11 +47,10 @@ class CommittableBatch {
 
     public synchronized int updateOffset(TopicPartition topicPartition, long offset) {
         if (this.outOfOrderCommits) {
-            List<Long> uncommittedThisTP = this.uncommitted.get(topicPartition);
+            SortedSet<Long> uncommittedThisTP = this.uncommitted.get(topicPartition);
             if (uncommittedThisTP != null && uncommittedThisTP.contains(offset)) {
-                List<Long> offsets = this.deferred.computeIfAbsent(topicPartition, tp -> new LinkedList<>());
-                if (!offsets.contains(offset)) {
-                    offsets.add(offset);
+                SortedSet<Long> offsets = this.deferred.computeIfAbsent(topicPartition, tp -> new TreeSet<>());
+                if (offsets.add(offset)) {
                     batchSize++;
                     this.inPipeline--;
                 }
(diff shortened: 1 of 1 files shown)


garyrussell:
Thanks, regarding the duplicates, I was concerned about user code, but now I see it is protected in `ConsumerHandler.CommittableOffset.maybeUpdateOffset()` so it looks to me like it would be safe to remove it.

pdanelson:
True. That check should cover the case of duplicates from user code acknowledging the same record object twice. However, is there also a possibility of another race condition that could trigger duplicates in the deferred queue?
1. A record is polled and handed over to the user code for processing (which may asynchronously take an arbitrarily long time)
2. A rebalance is triggered, forcing the revocation and re-assignment of partitions (and thus re-polling a new object of the same record that is still being processed by user code)
3. User code finally completes processing and acks the originally polled record object
4. User code completes processing and acks the second object of the same record that was re-polled, causing a duplicate offset in the deferred queue and thus halting further commit progress due to misalignment with the uncommitted queue

If the above scenario is already protected against in some manner (or I am misunderstanding some of the internals) and can not happen, then yeah, I can indeed modify this PR to simply remove the presence checks altogether.

garyrussell:
True, even though we now delay a rebalance, ostensibly until the pipeline is drained, the `maxDelayRebalance` could be exceeded causing the problem you describe. Before adding the rebalance delay, I considered adding en epoch to the `CommitableOffset`, but that could get complicated, especially with cooperative rebalancing; we'd have to maintain an epoch for each partition.

If your performance needs are satisfied with the current solution, let's go with it for safety.

pdanelson:
Safety first sounds good, especially since the performance difference isn't _too_ dramatic and this solution satisfies our needs by a fair margin. Thanks for the quick feedback!

garyrussell (approved):
Thanks for the contribution - looking forward to more.

Issue #309, opened 2023-01-13
Remove invasive error-logging in DefaultKafkaSender / SendProducer.java


kotlinlukas:
## Motivation
The DefaultKafkaSender fails with an error in console, if i.e. the record that we plan to send is too large (as intended).
However, when we catch and handle that error in our logic, we notice that the original error is still logged to the console as error, despite being caught and handled in our application logic.
This is causing problems for our logging and monitoring which notices errors in our log output, despite the fact that the application has not crashed and is handling the error just fine.

After some investigation, we noticed that the reason for this is the log.error statement in SendProducer.java: https://github.com/reactor/reactor-kafka/blob/main/src/main/java/reactor/kafka/sender/internals/SendSubscriber.java#L100

This is from the Reactor-Kafka internals, so there is no way for us to remove this log statement (we have now turned off kafka-logging entirely as a temporary workaround).

## Desired solution
Please remove this log statement (or log at debug/trace-level). An error is returned from this function anyways, so we can decide how to deal with the error case in our application code. There is no need imo to additionally log the error to console, at least not at error level, since it causes the aforementioned issues.

## Considered alternatives
Workaround is to disable kafka logging completely in config, but that's not great

kotlinlukas:
Hello, any update on this?

artembilan:
@kotlinlukas ,

Contribution is welcome!
Just open pull request with possible fix and we'll review it shortly.

Thanks

garyrussell:
Closed by dfa65c34b4c8a6212c1de619487bcec57b708a2c

PR #322, opened 2023-02-08
Fix invasive logging (#309)

kotlinlukas:
Fix for #309 
Changed log level to trace. Full reasoning in the issue, but in summary: the error is thrown anyways and can/should be managed/logged by the users of the library. The logging statement in the internal file was on `error` level, which is too high and caused issues in our monitoring system.

### DIFF
--- a/src/main/java/reactor/kafka/sender/internals/SendSubscriber.java
+++ b/src/main/java/reactor/kafka/sender/internals/SendSubscriber.java
@@ -97,7 +97,7 @@ class SendSubscriber<K, V, C> implements CoreSubscriber<ProducerRecord<K, V>> {
             }
 
             if (exception != null) {
-                DefaultKafkaSender.log.trace("Sender failed: $exception ");
+                DefaultKafkaSender.log.trace("Sender failed: $exception");
                 firstException.compareAndSet(null, exception);
                 if (senderOptions.stopOnError() || senderOptions.fatalException(exception)) {
                     onError(exception);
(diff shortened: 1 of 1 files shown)


src/main/java/reactor/kafka/sender/internals/SendSubscriber.java
```diff
@@ -97,7 +97,7 @@ public void onNext(ProducerRecord<K, V> record) {
             }
 
             if (exception != null) {
-                DefaultKafkaSender.log.error("Sender failed", exception);
+                DefaultKafkaSender.log.trace("Sender failed: $exception");
```
garyrussell, line 100:
```suggestion
                DefaultKafkaSender.log.trace("Sender failed: ", exception");
```

garyrussell (changes requested):
Also fix the copyright date in the file - see the PR build report.

`./gradlew spotlessApply`

src/main/java/reactor/kafka/sender/internals/SendSubscriber.java
artembilan: 😄 Right, this is not Kotlin

kotlinlukas:
The gradle task produces no changes in my project locally, but i've adjusted the copyright date based on the error message from the CI :)

garyrussell:
Oops - my suggestion had a syntax error; sorry about that.

src/main/java/reactor/kafka/sender/internals/SendSubscriber.java
```diff
@@ -97,7 +97,7 @@ public void onNext(ProducerRecord<K, V> record) {
             }
 
             if (exception != null) {
-                DefaultKafkaSender.log.error("Sender failed", exception);
+                DefaultKafkaSender.log.trace("Sender failed: ", exception");
```
artembilan, line 100:
```suggestion
                DefaultKafkaSender.log.trace("Sender failed: ", exception);
```

artembilan (changes requested):

garyrussell (approved):

Issue #321, opened 2023-02-07
Reactor Core + reactor-core-micrometer + context-propagation + tap, still no trace ID seen


patpatpat123:
Hello Reactor Kafka team,

I wanted to reach out reporting a small issue please:

I am having this very simple consumer based on Reactor Kafka:

```
package org.example;

import io.micrometer.observation.ObservationRegistry;
import org.apache.kafka.clients.consumer.ConsumerRecord;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.boot.CommandLineRunner;
import org.springframework.stereotype.Service;
import reactor.core.observability.micrometer.Micrometer;
import reactor.core.publisher.Flux;
import reactor.core.publisher.Mono;
import reactor.kafka.receiver.KafkaReceiver;
import reactor.kafka.receiver.ReceiverOptions;

import java.time.Duration;

@Service
public class MyConsumer implements CommandLineRunner {

    private static final Logger LOGGER = LoggerFactory.getLogger(MyConsumer.class);

    @Autowired
    private ReceiverOptions<String, String> receiverOptions;

    @Autowired
    private ObservationRegistry observationRegistry;


    @Override
    public void run(String... args) {
        myConsumer().subscribe();
    }

    public Flux<String> myConsumer() {
        return KafkaReceiver
                .create(receiverOptions)
                .receive()
                .concatMap(oneMessage -> consume(oneMessage), 500)
                .name("greeting.call")		//1
                .tag("latency", "low")	//2
                .tap(Micrometer.observation(observationRegistry));
    }

    private Mono<String> consume(ConsumerRecord<String, String> oneMessage) {
        LOGGER.info(String.format("<= Hope for traceID", Thread.currentThread().getName(), oneMessage));
        String transformedStringCPUIntensiveNonButNonBLocking = oneMessage.value().toUpperCase();
        return Mono.delay(Duration.ofSeconds(5L)).just(transformedStringCPUIntensiveNonButNonBLocking);
    }

}

```

Please note, in the code, you do see:

the pom:
```
<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xmlns="http://maven.apache.org/POM/4.0.0"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

[...]

PR #325, opened 2023-02-17
GH-321: Add Observation propagation

artembilan:
Fixes https://github.com/reactor/reactor-kafka/issues/321

* Copy-paste (and adjust) `KafkaReceiverObservation` & `KafkaSenderObservation` infrastructure from Spring for Apache Kafka
* Add an `Observation` handling into a `KafkaSender` with `contextCapture()` and `KafkaSenderObservation` around `producer.send()`
* An `Observation` handling on the consumer side is only around a consumer record to start and stop immediately to fill a gap for `COSNUMER` kind span in the propagated trace from producer.
If there is need to propagate an observation to downstream flow, then `KafkaReceiverObservation` API has to be used in the `flatMap()` around the `Mono` for record processing.
* The `ReactorKafkaObservationTests` demonstrates a single trace propagation from a sender via parent `Observation` and its restoration on a consumer side via `KafkaReceiverObservation` in the end-user code


artembilan:
/CC @marcingrzejszczak , @chemicL .

I believe it might not be possible to stick with an old Reactor version to preserve the current version.
So, I guess this really may lead to a new version for this project and respective upgrades for everything.

I converted this to `Draft` since I don't have yet a clear vision how to handle an `Observation` on a consumer side per record.
The current `KafkaReceiverObservation` is OK to restore the trace as a `CONSUMER` kind and have it connected to respective parent trace from the producer, but this is what end-user have to to ad the moment in their code.
I doubt we can propagate each record through Reactor context.

Thanks

src/main/java/reactor/kafka/sender/ImmutableSenderOptions.java
```diff
@@ -21,51 +21,88 @@
...
+        boolean userConfiguredClientId = this.properties.containsKey(ProducerConfig.CLIENT_ID_CONFIG);
+        if (userConfiguredClientId) {
+            refinedClientId = (String) this.properties.get(ProducerConfig.CLIENT_ID_CONFIG);
+        } else {
+            String transactionalId = (String) this.properties.get(ProducerConfig.TRANSACTIONAL_ID_CONFIG);
+            refinedClientId =
+                    "producer-" +
+                            (transactionalId != null ? transactionalId : PRODUCER_CLIENT_ID_SEQUENCE.getAndIncrement());
+        }
+        this.properties.put(ProducerConfig.CLIENT_ID_CONFIG, refinedClientId);
```
artembilan, line 105: I still have doubts about this thought, but we definitely need some identifier to attach to an `Observation` for traceability.

src/test/resources/log4j2-test.xml
```diff
@@ -17,7 +17,7 @@
 <Configuration status="WARN">
   <Appenders>
     <Console name="STDOUT" target="SYSTEM_OUT">
-      <PatternLayout pattern="[%d] [%t] [%c] %p %m%n" />
+      <PatternLayout pattern="[%d] [%t] [%c] %p [%X{traceId:-},%X{spanId:-}] %m%n" />
```
artembilan, line 20: I don't see anything in logs though.
Do I need to have something else for MDC to be populated automatically ?
For example some special `ObsrvationHandler`...

patpatpat123:
Cheering for this PR

src/test/resources/log4j2-test.xml
marcingrzejszczak: that should work oob

artembilan: Maybe it works with Spring Boot, but here I use a plain Brave and `ObservationRegistry` in the unit test and I have a Log4J and slf4J as deps for logging.
See `ReactorKafkaObservationTests` for more details.

artembilan: It just doesn't layout a simple observation in logs like this:
```
Observation.createNotStarted("test observation", OBSERVATION_REGISTRY)
                .observe(() -> log.trace("Current trace"));
```
🤷

artembilan: So, turns out I need to do this in my unit test:
```
        ThreadLocalCurrentTraceContext traceContext =
                ThreadLocalCurrentTraceContext.newBuilder()
                        .addScopeDecorator(MDCScopeDecorator.get())
                        .build();
```
Instead of plain `ThreadLocalCurrentTraceContext.create()`.
And now I can see a log correlation for my observations.

intosoft:
Hi, I think that tracing is still missing in 1.3.17-SNAPSHOT.

I managed to turn on traces in 1.3.16 by extending

ReactiveKafkaProducerTemplate:

`class ReactiveTracingKafkaProducerTemplate<K, V>(
    senderOptions: SenderOptions<K, V>,
    private val kafkaAdmin: KafkaAdmin?,
    private val observationRegistry: ObservationRegistry
) :
    ReactiveKafkaProducerTemplate<K, V>(senderOptions), BeanNameAware {

    private lateinit var beanName: String
    private var clusterId: String? = null

    override fun <T : Any?> send(record: SenderRecord<K, V, T>): Mono<SenderResult<T>> {
        val observation = KafkaTemplateObservation.TEMPLATE_OBSERVATION.observation(
            null, KafkaTemplateObservation.DefaultKafkaTemplateObservationConvention.INSTANCE,
            { KafkaRecordSenderContext(record, this.beanName) { this.clusterId() } },
            observationRegistry
        )

        return super.send(record).contextWrite(Function.identity())
            .doFirst { observation.start() }
            .doOnSuccess {
                observation.stop()
            }
            .doOnError {
                observation.error(it)
                observation.stop()
            }
    }

    @Nullable
    private fun clusterId(): String? {
        if (kafkaAdmin != null && this.clusterId == null) {
            this.clusterId = kafkaAdmin.clusterId()
        }
        return this.clusterId
    }

    override fun setBeanName(beanName: String) {
        this.beanName = beanName;
    }
}`

And ReactiveKafkaConsumerTemplate:

`
class ReactiveTracingKafkaConsumerTemplate<K, V>(
    receiverOptions: ReceiverOptions<K, V>,
    private val kafkaAdmin: KafkaAdmin?,
    private val observationRegistry: ObservationRegistry
) :
    ReactiveKafkaConsumerTemplate<K, V>(receiverOptions), BeanNameAware {
    private lateinit var beanName: String
    private var clusterId: String? = null

    override fun receiveAutoAck(): Flux<ConsumerRecord<K, V>> {
        return super.receiveAutoAck()
            .map {
                val observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
                    null,
                    DefaultKafkaListenerObservationConvention.INSTANCE,
                    {
                        KafkaRecordReceiverContext(
                            it, this.beanName
                        ) { clusterId() }
                    },
                    observationRegistry
                )
                observation.observe {  }
                it
            }.contextWrite(Function.identity())
    }

    @Nullable
    private fun clusterId(): String? {
        if (kafkaAdmin != null && this.clusterId == null) {
            this.clusterId = kafkaAdmin.clusterId()
        }
        return this.clusterId
    }

    override fun setBeanName(beanName: String) {
        this.beanName = beanName;
    }
}`

I know this solution is far from perfect one but maybe you could move some parts of this code to reactor kafka project.

artembilan:
@Tomasz-Marciniak ,

Thank you for sharing the code, but what you show so far has nothing to do with Reactor Kafka.
It is more Spring for Apache Kafka stuff, but when we come up with the fix for this project, we won't to do anything on Spring for Apache Kafka side.

Your producer side is similar to what I did in the `KafkaSender` code.
The consumer is a bit awkward since what you do is just observe an empty action: `observation.observe {  }`.
I'm not sure that this is what users would like to deal with.
Therefore in this PR I'm thinking about some callback for users to make something useful from an observation when you process a consumed record. Withing the context of a `Mono` per record.

I'm not sure in your `.contextWrite(Function.identity())` though. you don't put your observation into a context, so there won't be anything to propagate back to `ThreadLocal`.

It is also not fair to say `I think that tracing is still missing in 1.3.17-SNAPSHOT.`, since this PR has not been merged yet and respective GH issue is not closed.

Tomasz-Marciniak:
@artembilan 

Thank you for your comment. Much appreciated. I understand this is related to Spring for Apache Kafka but under the hood spring-kafka uses DefaultKafkaSender. Since I have no opportunity to adjust this code I had to make a workaround and I wrapped DefaultKafkaSender by:

      public class ReactiveKafkaProducerTemplate<K, V> implements AutoCloseable, DisposableBean {
      
	      private final **KafkaSender**<K, V> sender;

      }
      
I just wanted to share that in case of sender/producer it may work like this:

    val observation = KafkaTemplateObservation.TEMPLATE_OBSERVATION.observation(
                null, KafkaTemplateObservation.DefaultKafkaTemplateObservationConvention.INSTANCE,
                { KafkaRecordSenderContext(record, beanName) { this.clusterId() } },
                observationRegistry
            )
    
            return super.send(record).contextWrite(Function.identity())
                .doFirst { observation.start() }
                .doFinally { observation.stop() }
                .doOnError { observation.error(it) }

I'm aware that 1.3.17-SNAPSHOT is not finished, I just did not know if you were planning to add the support for tracing in this release and I wanted to tell that in current library I could not find it.

Regarding .contextWrite(Function.identity()), I was just playing with that function, the traces for receiving message are visible in Tempo but the trace id are not visible in logs so I tried everything. It helped me in case of Mongo reactive repos. To be honest I do not know how to wrap around the whole Flux/Mono on the consumer side to track receiving the message.

Tomasz-Marciniak:
@artembilan 

I have went through PRs and seems I get your point related to callback. I managed to do something like this and it works:

    fun <T> receiveAutoAckWithCallback(consumerCallbackFunction: Function<Flux<ConsumerRecord<K, V>>, Flux<T>>): Flux<T> {
        return super.receiveAutoAck().flatMap { record ->
            Mono.just(record)
                .transformDeferred { it ->
                    val observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
                        null,
                        KafkaListenerObservation.DefaultKafkaListenerObservationConvention.INSTANCE,
                        { KafkaRecordReceiverContext(record, beanName) { clusterId() } },
                        observationRegistry
                    )
                    observation.start();
                    
                    consumerCallbackFunction.apply(Flux.from(it))
                        .doOnTerminate(observation::stop)
                        .doOnError(observation::error)
                        .contextWrite { context -> context.put(ObservationThreadLocalAccessor.KEY, observation) };
                }
        }

To invoke it I simply use this:

    override fun consume(): Flux<UUID> {
          return template.receiveAutoAckWithCallback {
              receiveFluxCallback(it)
          }
      }

    private fun receiveFluxCallback(it: Flux<ConsumerRecord<String, String>>): Flux<UUID> =
        it
            .doOnNext { consumerRecord: ConsumerRecord<String, String> ->
                log.info(
                    "Received kafka message with attributes: key={}, value={} from topic={}, offset={}",
                    consumerRecord.key(),
                    consumerRecord.value(),
                    consumerRecord.topic(),
                    consumerRecord.offset()
                )
            }.map {
                UUID.randomUUID()
            }

I'm not sure the Function is the best approach here but for now, I'm ok with that and in the future, I'm sure I will have no such challenges with 1.3.17 :)

artembilan:
Sure! That's also possible. We can consider it as an API evolution: I believe my current low-level approach still can be used for regular `receiveAutoAck()`.
My point is that is not known how end-user would like to process his/her records within an observation. Therefore it might be still OK to give such a chance to do that `transformDeferred()` manually.

The discussion is not over yet: your feedback is really valuable.

Thank you!

Tomasz-Marciniak:
Thank you Artem,

Yes with transformDeffered all further actions were gone in trace graphs, like saving in mongo for instance.

I ended up with:

    fun <T> receiveAutoAckWithCallback(function: Function<Flux<ConsumerRecord<K, V>>, Flux<T>>): Flux<T> {
        return super.receiveAutoAck()
            .flatMap {
                val observation = KafkaListenerObservation.LISTENER_OBSERVATION.observation(
                    null,
                    KafkaListenerObservation.DefaultKafkaListenerObservationConvention.INSTANCE,
                    { KafkaRecordReceiverContext(it, beanName) { clusterId() } },
                    observationRegistry
                )
                observation.start();
                function.apply(Flux.just(it)).doOnTerminate(observation::stop)
                    .doOnError(observation::error)
                    .contextWrite { context -> context.put(ObservationThreadLocalAccessor.KEY, observation) };
            }
    }

This produce nice trace graph:

image

src/main/java/reactor/kafka/receiver/observation/KafkaRecordReceiverContext.java
```diff
@@ -0,0 +1,65 @@
...
+    public KafkaRecordReceiverContext(ConsumerRecord<?, ?> record, String clientId, String kafkaServers) {
+        super((carrier, key) -> {
+            Header header = carrier.headers().lastHeader(key);
+            if (header == null) {
+                return null;
+            }
+            return new String(header.value(), StandardCharsets.UTF_8);
+        });
+        setCarrier(record);
+        this.record = record;
```
maciej-gromul, line 48: Since you are calling setCarrier to set the carrier and it has public get method, wouldn't it be better to simply use that instead of creating another reference with this.record ? Get source can then be like this:

```java
public String getSource(){
  return getCarrier().topic();
}
```

src/main/java/reactor/kafka/receiver/observation/KafkaRecordReceiverContext.java
```diff
@@ -0,0 +1,65 @@
...
+        this.clientId = clientId;
```
maciej-gromul, line 49: For clientId and topic maybe we could leverage the context from inherited Observartion.Context ? It has a map with put/get methods. Later that context can be propagated to span tags/attributes, metrics labels. etc.

src/main/java/reactor/kafka/receiver/observation/ReceiverObservations.java
```diff
@@ -0,0 +1,81 @@
...
+ * These factories are needed if there is a requirements to process a record in a reactive manner within
+ * the mentioned consumer observation.
+ * Otherwise, a regular {@link KafkaReceiverObservation} API is enough to use
+ * in a {@link reactor.core.publisher.Flux#doOnNext(Consumer)} operator.
+ *
+ * @see KafkaReceiverObservation
+ */
+public final class ReceiverObservations {
+
+    public static <T> Mono<T> observe(Mono<T> userMono, ConsumerRecord<?, ?> record, String clientId,
```
maciej-gromul, line 55: Could we make this into a Function<Flux<T>, Flux<T>> return so that it's easier to use with transformDeferred ? or maybe some tapping like it's in reactor-core ? I believe currently it's recommended to use Flux...tap(Micrometer.observation(observationRegistry)), wouldn't it be nice to make use of same "technique" ?

src/main/java/reactor/kafka/receiver/observation/KafkaRecordReceiverContext.java
artembilan: I'm not sure in this question. See `DefaultKafkaReceiverObservationConvention`. That's exactly where we populate our own key-values. Yes, in the end everything ends up in the mentioned map, but how is that really convenient to deal with the `Map` API instead of our own getters for the mentioned convention?

src/main/java/reactor/kafka/receiver/observation/KafkaRecordReceiverContext.java
artembilan: Sure! We can do that, but it feels more like nit-picking than useful change.
In the end it is just pointer to the same object in the memory. Not sure what problem you see here...

src/main/java/reactor/kafka/receiver/observation/ReceiverObservations.java
artembilan: Well, I probably don't understand `tap()` at all. So, would be great if you share more info and your thoughts.
My idea with this function is to let end-user to observe every `ConsumerRecord` withing its own `Mono` and an `Observation` tied to its context. How is the could be possible with the whole `Flux`?
How `tap()` can help us here, please?

src/main/java/reactor/kafka/receiver/observation/KafkaRecordReceiverContext.java
maciej-gromul: Oh it's not really a problem of a kind "it will leak/blow up". I just point out that it's duplicated content/reference which might confuse someone in future for a second.

src/main/java/reactor/kafka/receiver/observation/KafkaRecordReceiverContext.java
maciej-gromul: I've missed that part where it adds those keyvalues to observation context. In that case can we add to the observation keyvalues the topic and partition ? It is useful sometimes to know the exact source of the event. Something similar is currently done by sleuth implementation for it's kafka instrumentations.

src/main/java/reactor/kafka/sender/ImmutableSenderOptions.java
maciej-gromul: I believe that this is used by kafka producer. So in that case each trace id should be separate for each element or extracted from parent context. When you create SenderOptions it's rather impossible to achieve that, you need to have it on the observation scope when it's being created for each event.

src/main/java/reactor/kafka/receiver/observation/ReceiverObservations.java
maciej-gromul: I'm not that proficient with it yet, but i've seen reactor team using it. If i understand it properly, they are using signallistener registered with tap to start observations on doFirst, doOnNext and closing them based on complete signal (doOnCancel, doOnComplete, doOnError etc.). But I didn't have time to properly analyse it all.

Here's example of their implementation:

https://github.com/reactor/reactor-core/blob/main/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/MicrometerMeterListenerConfiguration.java

https://github.com/reactor/reactor-core/blob/main/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/MicrometerObservationListener.java

src/main/java/reactor/kafka/receiver/observation/KafkaReceiverObservation.java
```diff
@@ -0,0 +1,104 @@
...
+     */
+    RECEIVER_OBSERVATION {
+        @Override
+        public Class<? extends ObservationConvention<? extends Context>> getDefaultConvention() {
+            return DefaultKafkaReceiverObservationConvention.class;
+        }
+
+        @Override
+        public String getPrefix() {
+            return "spring.kafka.receiver";
```
ffaoudi, line 46: Maybe "reactor.kafka.receiver" instead of "spring.kafka.receiver" ?

src/main/java/reactor/kafka/receiver/observation/ReceiverObservations.java
Tomasz-Marciniak: @maciej-gromul 

> Could we make this into a Function<Flux, Flux> return so that it's easier to use with transformDeferred ? or maybe some tapping like it's in reactor-core ? I believe currently it's recommended to use Flux...tap(Micrometer.observation(observationRegistry)), wouldn't it be nice to make use of same "technique" ?

Have you seen this post? It does exactly what you are mentioning:

https://github.com/reactor/reactor-kafka/pull/325#issuecomment-1460517792

maciej-gromul: @Tomasz-Marciniak I've seen that, I'm not that good with kotlin but I think this is similar to what i've written. Generally this is how it should look like if someone want's to instrument the tracing manually to the receivers. 

```java
receiver.receive()
  .tap(SomeObservationListener.observation())

//or

receiver.receive()
  .transformDeferred(tracingContextDeferMethodFromSomewhere)
```

So lessening the possible boilerplate for api clients. Maybe we could even add it to KafkaReceiver implementation as  part of ReceiverOptions ? That way someone would simply do:

```java
KafkaReceiver.create(ReceiverOptions.create().observable(SomeObservableConfiguration))
```

and then under the hood DefaultKafkaReceiver can simply check that configuration and if needed apply tracing transformation. If someone wants to have more control they can simply not register that and manually execute that tap/transformDeferred

artembilan: Thank you for feedback!

One question though: does it work per consumer record?
The idea is to be able to restore a trace from produce and apply it for just consumed record.
I'm not sure that in most case it would be just enough to record a span for consumed record and give up.
Probably end-user would like to continue a processing with that observation on a record and produce more spans downstream.

maciej-gromul: There's this implementation on reactor-core I find to be really nice [MicrometerObservationListener](https://github.com/reactor/reactor-core/blob/main/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/MicrometerObservationListener.java) that is simply used as 
```java
Flux.just(1,2,3)
  .name("something")
  .tag("foo", "bar")
  .tap(Micrometer.observation(registry))
```

Needs the least knowledge from api users and looks rather clean :P

maciej-gromul: > One question though: does it work per consumer record?

I'm using transformDeferred with sleuth reactor on spring boot 2, and it works with brave instrumentation. I'm not sure about tap in spring boot 3 + micrometer. I didn't get to test it because for me current observations context doesn't work properly even with standard publishers, for some reason spans end up not propagating the context correctly downstream.

patpatpat123:
Just checking on this PR, it is a great initiative, hope there is nothing blocking it!

src/main/java/reactor/kafka/receiver/observation/ReceiverObservations.java
artembilan: OK. Sorry for delay: was busy with other stuff.
So, I read a bit about `tap()` and saw how it works in the source code.
My conclusion that it has very minimal functionality for our use-case.
And it is here like this: the `ConsumerRecord` has an info about a trace. 
It would be great to be able to restore that trace in the consumer and start a new child span.

The `tap()` wants from us to do something on subscription to the `Publisher`. 
For example, the mentioned `Micrometer.observation()` does start an `Observation` at that point.
All the upstream operators in the `Flux/Mono` will be aware of that observation then.
This observation is stopped in the `doOnNext` for `Mono` and in the `doOnComplete` for `Flux`.
Therefore everything after the `tap()` is not aware about that observation already.

If you look into my `ConsumerRecord` observation requirement, you'll see that this `tap()` won't fit.
Sure, we probably can implement something like "just restore and record" hook, then a simple `doOnNext` for that `tap` listener would be enough.
But is it a goal we would try to reach here? The observation will be restored from the record, started and stopped immediately. Just because this is a nature how the `tap()` works.
It is probably going to be fully similar what we have with a `handle()`. Therefore no reason to abuse `tap()` where we won't get any gains from its purpose.

So, what do I miss with the `tap()`, please?


Sounds like `transformDeferred` is still only the way for us to be able to restore such an observation from a consumer record and let it to be spread around some end-users operator. 
We might be need to think about eliminating an extra `ConsumerRecord` argument for that function and something like `Flux` input style would be great as well.
But this is something what remains to be explored.

Thank you for understanding!

maciej-gromul: I've had some issues created for observations with tap and like you say it works only in a context of whole publisher while propagating the context only downstream. I've created suggestion/request issue to give some kind of ability to publish the context upstream like it was done with spring sleuth but right now it's impossible. Current usage of observations in case of flux that should behave like a mono requires either having the main contract as Flux<Mono<T>>receive(), then each mono will have it's own context that consumers can read in their flatmaps or having callable for either flatmap / transform to execute, both changing the standard way reactor pipelines are written, forcing unintuitive schemas in case we ever need telemetry.

maciej-gromul: Potentially we could write custom implementation `KafkaSubscriber implements InnerOperator<T, T>` that could for each element read from it's internal source change the context available in methods onNext/tryOnNext and handle errors with onError/cancel/onComplete. But that would need contribution on ReactorCore since InnerOperator is package scoped.

patpatpat123: Fingers crossed for this

artembilan:
> Would it be possible to merge this?

I didn't come up with clean end-user API for Consumer observation.
Plus I think now that it is better to ask for a client id from end-user when we user observation instead of trying to infer it from producer props.

tongdaoqa:
Was about to start a feature request for the same. Found this (long opened) thread.

Big +1 for this!

What is blocking the merge please?
Non reactive have tracing well in place

kuldeep0508:
Any update on this MR ?

yusuf-murodov-epam:
Hi @artembilan, any chance to merge this in near future?

Thanks

patpatpat123:
This is great! Thanks @artembilan ! Hope this can be merged!

garyrussell (commented):
LGTM - any thoughts about adding documentation?

We can merge this for next week's release, if you prefer, and add docs later?

chemicL:
I had a look. It looks like this is an important improvement for the lives of developers :) I have a few questions/suggestions to understand the implementation and the intentions:
- The default pattern is to create an artificial `Observation` for each item in the `Flux` returned via `receive()` and immediately close the `Observation` – therefore any timing information is insignificant (both for metrics and for tracing data)?
- Assuming the above is true, the user also won't get any log correlation out of the change, unless they manually handle each item via a `flatMap`-like operator, where each item gets its own `Mono`, to which the `Observation` inferred is attached, correct?
- Docs for this would be really appreciated I suppose.
- Any chance to enhance the API to save the users from doing the manual work? E.g. `receiveObserved` variants, which could accept an item transformer (`Function<ReceiverRecord, Mono<T>>`), which would get the inferred `Observation` attached via contextWrite to the chain that the user provides?

artembilan:
Thank you, @chemicL , for review!
Here are answers to your questions:

> The default pattern is to create an artificial Observation for each item in the Flux returned via receive()

Well, that's correct, since there is no composition with end-user code and there is no way to guess if end-user is going to deal with an observation in their own code, we suggest that one-tick hook just to attach a `CONSUMER` span to the trace with that observation on each record.
Right, the metric will be void since we don't do any operations around consumed record. But again: this is just for convenience if you'd like to have a whole trace from producer to consumer on specific record.

> the user also won't get any log correlation out of the change

I believe we get those correlations only if we really produce logs.
Even for producer where we wrap its `send()` with an observation, we will get that trace info in logs only if we enable `trace` for a `org.apache.kafka.clients.producer` category.
I can suggest something like logging this received record within that observation at `debug` level.

> Docs for this would be really appreciated I suppose.

Yes, docs is the next step in this work. 
I assume that I cannot make it for tomorrow's release, so we may just delay this PR for a little more.

> Any chance to enhance the API to save the users from doing the manual work?

Yes, this is the one which made me stuck for the progress.
Now I believe we can think about it afterwards: in the end there is nothing wrong with using an `Observation` API manually in those `flatMap()` operations the user would compose to the result of the current `receive()`.
The point is that this is not a WebFlux handler or RSocket responder plugged to some infrastructure. 
We just provide a template-like API where you are on your own with the result.

If these all makes sense, I'll go ahead with docs whenever I can.
Something like `receiveAndHandle(Function)` variants could be introduced later on.

patpatpat123:
Would it be possible to have this merged as part of a snapshot build so we can start testing?

We already have ready a scenario where we have multiple producers, sending messages to kafka. Then, a smaller set of consumers consuming those messages (higher load here). And the consumers will send http requests to another service // save into db // send to yet another consumer. 

All in different load ranging from gentle to hardcore. 

We expect traces to flow all the way, having the possibility to search them in our Splunk, Zipkin, etc.

We would like to test this and provide feedback. Maybe some sort of snapshot where we can just test (even without doc) would
 be great

chemicL:
Thanks for explaining. I noticed one more thing. Is the fact that reactor-kafka depends on reactor-core 3.4.x, which doesn't provide any support for Micrometer Observation any issue here? Will you suggest users in the docs to ensure reactor-core 3.5.x is used to provide proper integration between reactor and micrometer?

Regarding the timing, I think it's safer to make the release today without this feature, but it can be merged afterwards to provide the snapshots and the ability for testing it with the chance to provide any improvements for the next weeks before it gets released. Then @patpatpat123 can have a go with a snapshot and let us know how it behaves :)

artembilan:
That's correct, @chemicL , there is no Reactor `3.5.x` changes in the project to start a new version and this indeed works with the previous one even if there is no convenient API for observation.
But yes: the recommendation is definitely use the latest Reactor for better developer and production experience.

I just don't want to start a new version here since it would cause some burden to support several branches.

michalwilk96:
Hello,
@artembilan, @garyrussell could you please make a higher priority on this PR? 
Thank you in advance

patpatpat123:
Gentle bump

garyrussell:
@artembilan When you have time, please rebase and resolve conflicts.

artembilan:
Sure! Will do today/tomorrow.
Do you want me to add some docs on the matter?

garyrussell:
Yes; that would be good; thanks.

src/docs/asciidoc/api-guide.adoc
```diff
@@ -608,3 +608,39 @@ methods can be invoked after the receive Flux corresponding to the last receive
 
 To enable micrometer metrics for the underlying Kafka Consumers and Producers, add a `MicrometerConsumerListener` to the `ReceiverOptions` or a `MicrometerProducerListener` to the `SenderOptions` respectively.
 
+=== Micrometer Observation
+
+To enable Micrometer observations for produced and consumed records, add an `ObservationRegistry` to the `SenderOptions` and `ReceiverOptions` using `withObservation()` API.
```
garyrussell, line 613:
```suggestion
To enable Micrometer observation for produced and consumed records, add an `ObservationRegistry` to the `SenderOptions` and `ReceiverOptions` using the `withObservation()` API.
```

src/docs/asciidoc/api-guide.adoc
```diff
@@ -608,3 +608,39 @@ methods can be invoked after the receive Flux corresponding to the last receive
...
+A custom `KafkaSenderObservationConvention` (and `KafkaReceiverObservationConvention`) can also be set.
+See their default implementations in the `KafkaSenderObservation` and `KafkaReceiverObservation`, respectively.
+The `DefaultKafkaSenderObservationConvention` exposes two low-cardinality tags: `reactor.kafka.type = sender`  and `reactor.kafka.client.id` with the `ProducerConfig.CLIENT_ID_CONFIG` option or identity hash code of the `DefaultKafkaSender` instance prefixed with the `reactor-kafka-sender-`.
+The `DefaultKafkaReceiverObservationConvention` exposes two low-cardinality tags: `reactor.kafka.type = receiver`  and `reactor.kafka.client.id` with the `ConsumerConfig.CLIENT_ID_CONFIG` option or identity hash code of the `DefaultKafkaReceiver` instance prefixed with the `reactor-kafka-receiver-`.
+
+If a `PropagatingSenderTracingObservationHandler` is configured on the `ObservationRegistry`, the tracing information from the context around producer record is stored into its headers before publishing this record to Kafka topic.
```
garyrussell, line 619:
```suggestion
If a `PropagatingSenderTracingObservationHandler` is configured on the `ObservationRegistry`, the tracing information from the context around a producer record is stored into its headers before publishing this record to the Kafka topic.
```

src/docs/asciidoc/api-guide.adoc
```diff
@@ -608,3 +608,39 @@ methods can be invoked after the receive Flux corresponding to the last receive
...
+If a `PropagatingReceiverTracingObservationHandler` is configured on the `ObservationRegistry`, the tracing information from the mentioned Kafka record headers, is restored into the context on a receiver side with a child span.
```
garyrussell, line 620:
```sugggestion
If a `PropagatingReceiverTracingObservationHandler` is configured on the `ObservationRegistry`, the tracing information from the mentioned Kafka record headers, is restored into the context on the receiver side with a child span.
```

src/docs/asciidoc/api-guide.adoc
```diff
@@ -608,3 +608,39 @@ methods can be invoked after the receive Flux corresponding to the last receive
...
+
+Because the reverse order nature of the Reactor context, an observation functionality on the `KafkaReceiver` is limited just to a single `trace` logging message for received record.
```
garyrussell, line 622:
```suggestion
Because the reverse order nature of the Reactor context, the observation functionality on the `KafkaReceiver` is limited just to a single `trace` logging message for each received record.
```

src/docs/asciidoc/api-guide.adoc
```diff
@@ -608,3 +608,39 @@ methods can be invoked after the receive Flux corresponding to the last receive
...
+A restored tracing information will be correlated into logs if configured so for logging system.
+If there are requirements to continue an observation on the consumer side, the `KafkaReceiverObservation.RECEIVER_OBSERVATION` API must be used manually in the record processing operator:
```
garyrussell, line 624:
```suggestion
Restored tracing information will be correlated into logs if so configured for the logging system.
If there are requirements to continue an observation on the consumer side, the `KafkaReceiverObservation.RECEIVER_OBSERVATION` API must be used manually in the record processing operator:
```

garyrussell (changes requested):

garyrussell (approved):

patpatpat123:
This is great! Thanks! 

I would like to help test this. Just to be sure I am testing the "right thing"

Should I pull the latest available snapshot from https://repo.spring.io/snapshot ?

michalwilk96:
Hello @garyrussell, @artembilan,
when do you plan the next reactor-kafka release?

garyrussell:
November 7: https://calendar.spring.io/

artembilan:
@patpatpat123 ,

yes, the latest SNAPSHOT is available in that repo.

michalwilk96:
Hi @artembilan , @garyrussell 
I don't see that fix on the list of reactor-kafka 1.3.22 release: https://github.com/reactor/reactor-kafka/milestone/39?closed=1. Is it true?

michalwilk96:
> November 7: https://calendar.spring.io/

@garyrussell 7 November or 14?

jonatan-ivanov:
@michalwilk96 Adding the milestone to the issue and to the PR as well would show up as a duplicate so usually the milestone is added to the issue: https://github.com/reactor/reactor-kafka/issues/321

artembilan:
Just added that issue to the respective milestone.
Yes, November 14: https://github.com/reactor/reactor-kafka/milestone/39

patpatpat123:
Hello team,

Just verified 1.3.22 was not part of SpringBoot 3.2.0-RC2

Would it be possible to have this PR as part of the official 3.2.0 release due one week after 1.3.22 of reactor kafka?
https://github.com/spring-projects/spring-boot/milestone/308?closed=1

Thank you

artembilan:
This is release is going to be a part of upcoming Reactor release trains tomorrow.
And indeed those Reactor BOMs are going to be pulled automatically by upcoming Spring Boot releases next week.

violetagg:
> @michalwilk96 Adding the milestone to the issue and to the PR as well would show up as a duplicate so usually the milestone is added to the issue: #321

That's not how we do this in Project Reactor. It is important that PRs has the milestone and the type, because we need to generate the release notes based on this information.

mukul20798:
Hi @marcingrzejszczak , [Here](https://github.com/reactor/reactor-kafka/blob/c35681acdc05a21b1f76c57589957c44720f80c8/src/main/java/reactor/kafka/receiver/ReceiverOptions.java#L573) in this change, type casting(String) is used for **bootstrap.severs** value and when I am passing my server values(e.g listOf(val1,val2) as mentioned [here](https://github.com/spring-projects/spring-boot/blob/7216d2b031be648763744a371b871fa7c320c722/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java#L71) in springframework, then it is giving type cast exception as below



`java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class java.lang.String (java.util.ArrayList and java.lang.String are in module java.base of loader 'bootstrap')
	at reactor.kafka.receiver.ReceiverOptions.bootstrapServers(ReceiverOptions.java:573)
	at reactor.kafka.receiver.internals.DefaultKafkaReceiver.lambda$observerRecord$21(DefaultKafkaReceiver.java:160)
	at io.micrometer.observation.Observation.createNotStarted(Observation.java:169)
	at io.micrometer.observation.docs.ObservationDocumentation.observation(ObservationDocumentation.java:188)
	at reactor.kafka.receiver.internals.DefaultKafkaReceiver.observerRecord(DefaultKafkaReceiver.java:156)
	at reactor.core.publisher.FluxPeekFuseable$PeekFuseableSubscriber.onNext(FluxPeekFuseable.java:196)
	at reactor.core.publisher.FluxFlattenIterable$FlattenIterableSubscriber.drainAsync(FluxFlattenIterable.java:453)
	at reactor.core.publisher.FluxFlattenIterable$FlattenIterableSubscriber.drain(FluxFlattenIterable.java:724)
	at reactor.core.publisher.FluxFlattenIterable$FlattenIterableSubscriber.onNext(FluxFlattenIterable.java:256)
	at reactor.core.publisher.FluxPublishOn$PublishOnSubscriber.runBackfused(FluxPublishOn.java:484)
	at reactor.core.publisher.FluxPublishOn$PublishOnSubscriber.run(FluxPublishOn.java:521)
	at reactor.core.scheduler.ExecutorScheduler$ExecutorTrackedRunnable.run(ExecutorScheduler.java:192)
	at reactor.core.scheduler.ImmediateScheduler$ImmediateSchedulerWorker.schedule(ImmediateScheduler.java:84)
	at reactor.core.scheduler.SingleWorkerScheduler.execute(SingleWorkerScheduler.java:64)
	at reactor.core.scheduler.ExecutorScheduler$ExecutorSchedulerWorker.schedule(ExecutorScheduler.java:252)
	at reactor.core.publisher.FluxPublishOn$PublishOnSubscriber.trySchedule(FluxPublishOn.java:312)
	at reactor.core.publisher.FluxPublishOn$PublishOnSubscriber.onNext(FluxPublishOn.java:237)
	at reactor.core.publisher.FluxPeekFuseable$PeekFuseableSubscriber.onNext(FluxPeekFuseable.java:210)
	at reactor.core.publisher.SinkManyUnicast.drainRegular(SinkManyUnicast.java:283)
	at reactor.core.publisher.SinkManyUnicast.drain(SinkManyUnicast.java:364)
	at reactor.core.publisher.SinkManyUnicast.tryEmitNext(SinkManyUnicast.java:238)
	at reactor.core.publisher.SinkManySerialized.tryEmitNext(SinkManySerialized.java:100)
	at reactor.core.publisher.InternalManySink.emitNext(InternalManySink.java:27)
	at reactor.kafka.receiver.internals.ConsumerEventLoop$PollEvent.run(ConsumerEventLoop.java:380)
	at io.micrometer.context.ContextSnapshot.lambda$wrap$0(ContextSnapshot.java:91)
	at reactor.core.scheduler.SchedulerTask.call(SchedulerTask.java:68)
	at reactor.core.scheduler.SchedulerTask.call(SchedulerTask.java:28)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)`

artembilan:
@mukul20798 ,

this has nothing to do with observation support.
That is how Reactor Kafka was designed originally for that property.

we cannot change  that to realign with Spring Boot option in the current patch version.

here is how you have to configure for now on:
```
producerProperties.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG,
				StringUtils.collectionToCommaDelimitedString(kafkaProperties.getBootstrapServers()));
```
Feel free to raise a new GH issue, so we will review this when we start a new version.

and stop, please, comment on old closed issues/PRs: the comment might not be related to the subject or easily lost. Because , you know , it is closed already .

PR #359, opened 2023-08-31
Add receive batch to receiver

### LINKED ISSUE #261: receiveAutoAck commits the batch even in case of exception (or Error) while processing
[receiveAutoAck](https://github.com/reactor/reactor-kafka/blob/244fba511ab3f81c8699ce6d0a483cf929908447/src/main/java/reactor/kafka/receiver/KafkaReceiver.java#L111) returns a `Flux<Flux<Record>>` for batches of received records. The internal flux gets terminated (even in case of exception) and commits the batch, because of [doOnTerminate](https://github.com/reactor/reactor-kafka/blob/244fba511ab3f81c8699ce6d0a483cf929908447/src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java#L73) callback.

I might be missing something, but I don't see a way to avoid committing the batch if a crash occurs (such as OutOfMemory). 

I've prepared a [test case](https://gist.github.com/serejke/17fb3296db26b360bf92d3b8cf56c983)

`receiveAutoAck` is [said](https://github.com/reactor/reactor-kafka/blob/main/src/docs/asciidoc/api-guide.adoc) to be suitable for at-least-once delivery of messages, but in fact the events might be committed before they are processed.

mohamed-gara:
This PR is an implementation to resolve #261. I have chosen the name proposed by @serejke [here](https://github.com/reactor/reactor-kafka/issues/261#issuecomment-1011481582) for the new method in KafkaReceiver.

Actually, in our product I use receive() with the bufferTimeout operator. But it is not the best solution to consume a record batch. With the receiveBatch() method batch consumption will be easier.

### DIFF
--- a/src/main/java/reactor/kafka/receiver/KafkaReceiver.java
+++ b/src/main/java/reactor/kafka/receiver/KafkaReceiver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


src/main/java/reactor/kafka/receiver/KafkaReceiver.java
```diff
@@ -97,6 +97,35 @@ default Flux<ReceiverRecord<K, V>> receive() {
...
+     * Returns a {@link Flux} containing each batch of consumer records returned by {@link Consumer#poll(long)}.
+     * The maximum number of records returned in each batch can be configured on {@link ReceiverOptions} by setting
+     * the consumer property {@link ConsumerConfig#MAX_POLL_RECORDS_CONFIG}. Each batch is returned as one Flux.
+     * Every record must be acknowledged using ReceiverOffset.acknowledge() in order to commit the offset
+     * corresponding to the record. Acknowledged records are committed based on the configured commit interval
+     * and commit batch size in ReceiverOptions. Records may also be committed manually using ReceiverOffset.commit().
+     *
+     * @param prefetch amount of prefetched batches
+     * @return Flux of consumer record batches from Kafka that are committed only after acknowledgement
+     * @since 1.3.?
```
artembilan, line 110:
```suggestion
     * @since 1.3.21
```

src/main/java/reactor/kafka/receiver/KafkaReceiver.java
```diff
@@ -97,6 +97,35 @@ default Flux<ReceiverRecord<K, V>> receive() {
...
+    /**
+     * Returns a {@link Flux} containing each batch of consumer records returned by {@link Consumer#poll(long)}.
+     * The maximum number of records returned in each batch can be configured on {@link ReceiverOptions} by setting
+     * the consumer property {@link ConsumerConfig#MAX_POLL_RECORDS_CONFIG}. Each batch is returned as one Flux.
+     * Every record must be acknowledged using ReceiverOffset.acknowledge() in order to commit the offset
+     * corresponding to the record. Acknowledged records are committed based on the configured commit interval
+     * and commit batch size in ReceiverOptions. Records may also be committed manually using ReceiverOffset.commit().
+     *
+     * @return Flux of consumer record batches from Kafka that are committed only after acknowledgement
+     * @since 1.3.?
```
artembilan, line 123:
```suggestion
     * @since 1.3.21
```

artembilan (changes requested):
Just a couple `@since` correction.
Otherwise I don't see any problems with this fix.

Thanks

src/main/java/reactor/kafka/receiver/KafkaReceiver.java
mohamed-gara: Done.

src/main/java/reactor/kafka/receiver/KafkaReceiver.java
mohamed-gara: The two @since are fixed.

artembilan (approved):

garyrussell (changes requested):
LGTM; please add some documentation for the `receiveBatch` method.

https://projectreactor.io/docs/kafka/release/reference/#api-guide-receiver

artembilan (approved):

garyrussell (approved):

PR #258, opened 2021-12-22
Support Exactly Once Semantics (EOS Mode BETA V2)

garyrussell:
Resolves https://github.com/reactor/reactor-kafka/issues/207

Use `sendOffsetsToTransaction` that takes `ConsumerGroupMetadata` instead of
just he consumer `group.id`.

`kafka-clients` will currently downgrade to `ALPHA` (V1) if the broker is
earlier than 2.5.

PR #364, opened 2023-10-18
Draft for an exactly once processor with support for horizontal scaling

jrhino97:
As I understand it, for exactly once Zombie Fencing to function correctly with horizontally scaled processors, it has to be ensured that the Sender with the same transactionalId always consumes from the same inbound topic partitions. 

As the transactionalId has to be unique, I thought it might be a good idea, to create a Sender for each assigned Inbound Topic and have the TopicPartition be part of the Senders transactionalId. I read somewhere that Spring Kafka does it that way.  

I wanted to put my first (naive) implementation of this up for discussion, to see if I'm on the right track and if this is a feature that might be integrated into Reactor-Kafka.

Obviously, it still need refactoring and tests. But I wanted to discuss if I'm on the right track before moving further. 

Thank you for your consideration.

### DIFF
--- a/src/main/java/reactor/kafka/receiver/ExactlyOnceProcessor.java
+++ b/src/main/java/reactor/kafka/receiver/ExactlyOnceProcessor.java
@@ -1,0 +1,24 @@
+package reactor.kafka.receiver;
+
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Flux;
+import reactor.kafka.receiver.internals.DefaultExactlyOnceProcessor;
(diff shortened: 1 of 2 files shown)


garyrussell:
>t has to be ensured that the Sender with the same transactionalId always consumes from the same inbound topic partitions.

That is no longer the case; EOS mode BETA (aka V2) now uses consumer metadata when sending offsets to transactions.

See KIP-447

This project has supported the new mechanism since January 2022 - see https://github.com/reactor/reactor-kafka/pull/258

PR #366, opened 2023-10-26
Fix broken example code ('retry(When' -> 'retryWhen')

acktsap:
It seems example code is broken.

Screenshot 2023-10-26 at 5 24 19 PM


artembilan (approved):

artembilan:
Thank you for contribution; looking forward for more!

Issue #378, opened 2024-01-03
Provide option for KafkaReceiver's graceful shutdown


cjlee38:
## Motivation

Over the past few days, I've been looking for a way to shutdown gracefully KafkaReceiver, but couldn't find any proper way to handle this. I read related issues (https://github.com/reactor/reactor-kafka/issues/247, https://github.com/reactor/reactor-kafka/issues/51, https://github.com/reactor/reactor-kafka/issues/196) or SO questions but they don't work as I expected (This might be caused by my bad understanding of reactor or kafka, so please excuse me)

Example code snippets :

```kotlin
val disposable = kafkaReceiver.receive()
    .flatMapSequential { record -> process(record).thenReturn(record) }
    .concatMap { record -> 
        record.receiverOffset().acknowledge()
        record.receiverOffset().commit() 
    }
    .subscribe()
```

This is typical case.
1. receive record
2. process record
3. ack and commit (can be omitted when using auto-commit)

## Desired solution

So, I think this is very common case : when I re-start my application(which is based on spring framework), consumers stop fetching records, and ongoing(I mean, already fetched records) flux keeps processing and also commits, and then complete the flux.

However, just disposing the `disposable` would not work as expected, because there is possibility that processed record not be committed. 

## Considered alternatives

There is no concrete idea to implement this, but things to consider are next.

1. The `Scheduler` interface of reactor provides `disposeGracefully` method. 
   <img width="668" alt="image" src="https://github.com/reactor/reactor-kafka/assets/52564093/716af658-b56f-48c3-ba48-b8c62abb71e7">
   These methods ([1](https://github.com/reactor/reactor-kafka/blob/57b7085f81def57838ee992d4157a6493cee7dac/src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java#L191) [2](https://github.com/reactor/reactor-kafka/blob/57b7085f81def57838ee992d4157a6493cee7dac/src/main/java/reactor/kafka/receiver/internals/ConsumerHandler.java#L132)) can be replaced with this (or selected by option)
2. add `sink.emitComplete()`  in [ConsumerEventLoop#stop](https://github.com/reactor/reactor-kafka/blob/57b7085f81def57838ee992d4157a6493cee7dac/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java#L201)
[...]

Issue #247, opened 2021-10-05
Add JVM shutdown hook to KafkaReceiver to close KafkaConsumer


SujanKumarMitra:
When the JVM receives a SIGINT, 
the `KafkaReceiver` does NOT close the underlying `KafkaConsumer` to prevent resource leaks.

simonbasle:
Is that a typical arrangement @garyrussell @artembilan ?
If the `kafka-client` (`KafkaConsumer`) doesn't install such a hook and puts the responsibility on the user, doesn't it make sense that the same would be required for `KafkaSender` (and its `#close()` method)?

artembilan:
Well, Spring for Apache Kafka project is fully rely on the `ApplicationContext` lifecycle.
So, whatever calls its `close()` would initiate the underlying `KafkaConsumer.close()`.

Doesn't look like Spring adds such a hook by default: https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-factory-shutdown.
But at least there is such a API for end-user consideration.

Looks like Spring Boot has its own opinion: https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java#L203.

But anyway. If we expose something like `registerShutdownHook()` on the `KafkaReceiver` and `KafkaSender`, that should be enough for end-user requirements.

SujanKumarMitra:
I am using Reactor Kafka in Spring Boot Webflux environment. 

So, I am expecting that spring-managed `KafkaReceiver` beans should be closed on SIGINT. Like it happens for JDBC or MongoDB connections. 

But, Spring only closes for `KafkaSender`. Maybe if KafkaReceiver implements the `Closeable` interface or `close()`, spring's shutdown hook might kick in.

artembilan:
OK. I see. Right, the `KafkaReceiver` (neither its impl) has a `close()` API, unlike `KafkaSender` does.

At the same time I see that `ConsumerHandler` has one, but does not delegate to its `Consumer.close()`.

SujanKumarMitra:
I have a question. When exactly does the `KafkaReceiver.receive()` completes? I think, it runs infinitely unless someone uses the `Disposable.dispose()` method.

My proposal is,
expose a close() method in `KafkaReceiver` which will complete the `KafkaReceiver.receive()` flux and close the `KafkaConsumer` as well.

[...]

PR #380, opened 2024-01-17
Expose manual stop method in KafkaReceiver

cjlee38:
KafkaReceiver currently is not possible to close safely. An example is below:

```kotlin
val disposable = kafkaReceiver.receive()
    .flatMapSequential { record -> process(record).then(record) }
    .delayUntil { 
        it.receiverOffset().acknowledge() 
        it.receiverOffset().commit()
}
    .subscribe()
```

`dispose()` method would be a way,  but it might lead to unexpected result because of cancellation leading processed records not to be committed. I think This PR is discussed at #247, and also helps to address #378 indirectly.(People can customize KafkaReceiver's lifecycle after stopping followed by `disposable.isDisposed()`

I've considered several situations related to thread-safety or something, but might miss I didn't expect. Any further suggestions would be appreciated.

### DIFF
--- a/src/main/java/reactor/kafka/receiver/KafkaReceiver.java
+++ b/src/main/java/reactor/kafka/receiver/KafkaReceiver.java
@@ -283,4 +283,12 @@ public interface KafkaReceiver<K, V> {
      * @return Mono that completes with the value returned by <code>function</code>
      */
     <T> Mono<T> doOnConsumer(Function<Consumer<K, V>, ? extends T> function);
+
+    /**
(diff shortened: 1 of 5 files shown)


cjlee38:
Hello,

It has been quite some time since I submitted this PR, and I haven't received any response yet. I was wondering if support for reactor-kafka has ended, or if there might be another reason for the delay? I don't mean to rush, but I'm leaving this message because it has been a considerable amount of time without any reply.

Thank you.

Issue #371, opened 2023-11-13
Rebalancing always waits until maxDelayRebalance with AckMode.EXACTLY_ONCE


damn1kk:
Hello, I have a topic with multiple partitions and multiple app instances. I use AckMode.EXACTLY_ONCE with transactional sender.
Rebalance always waits until maxDelayRebalance.
I get a repeated message in the log: `reactor.kafka.receiver.internals.ConsumerEventLoop - Rebalancing; waiting for N records in pipeline`
Similar to the issue discussed [here in the comments  ](https://github.com/reactor/reactor-kafka/issues/345), but I have AckMode.EXACTLY_ONCE.

I think the problem is that the `inPipeline` variable changes in different objects.

`isPipeline` decreases in object created here:
https://github.com/reactor/reactor-kafka/blob/5436ae9063349432ceeff4caa18b14d953989526/src/main/java/reactor/kafka/receiver/internals/DefaultKafkaReceiver.java#L133-L135

but increases in another object here:
https://github.com/reactor/reactor-kafka/blob/5436ae9063349432ceeff4caa18b14d953989526/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java#L377

So variable `inPipeline` is always positive and the condition here is always waits until maxDelayRebalance:

https://github.com/reactor/reactor-kafka/blob/5436ae9063349432ceeff4caa18b14d953989526/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java#L170-L184

## Expected Behavior
Rebalancing should complete after all read messages have been processed and the transaction has committed.

## Actual Behavior
Rebalancing always waits until maxDelayRebalance.

## Example to Reproduce

https://github.com/damn1kk/kafka-rebalance-issue/blob/master/src/test/java/KafkaRebalanceTest.java
```
        ReceiverOptions<String, String> receiverOptions = ReceiverOptions.<String, String>create(
                        Map.of(
                                ConsumerConfig.CLIENT_ID_CONFIG, UUID.randomUUID().toString(),
                                ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafka.getBootstrapServers(),
                                ConsumerConfig.GROUP_ID_CONFIG, "consumer-group",
                                ConsumerConfig.MAX_POLL_RECORDS_CONFIG, 10,
                                ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "latest",
[...]

PR #381, opened 2024-01-21
fixed waiting for maxDelayRebalance with AckMode.EXACTLY_ONCE during rebalancing #371

damn1kk:
Resolves https://github.com/reactor/reactor-kafka/issues/371

Fixed waiting for maxDelayRebalance with AckMode.EXACTLY_ONCE during rebalancing.

Don't increment uncommitted messages in CommitBatch for AckMode.ExactlyOnce.
Increasing the number of uncommitted messages in CommitBatch causes an infinite loop during rebalancing because they never decrease for AckMode.ExactlyOnce.


damn1kk:
@violetagg , @chemicL , @artembilan 

can someone of you please have a look ?

PR #387, opened 2024-03-19
Start producing 1.4 snapshots with 1.4.0-SNAPSHOT

chemicL:


README.md
```diff
@@ -45,8 +45,8 @@ With Gradle from repo.spring.io:
     }
 
     dependencies {
-      compile "io.projectreactor.kafka:reactor-kafka:1.3.23"
-      // compile "io.projectreactor.kafka:reactor-kafka:1.3.24-SNAPSHOT"
+      // compile "io.projectreactor.kafka:reactor-kafka:1.4.1-SNAPSHOT"
+      compile "io.projectreactor.kafka:reactor-kafka:1.4.0"
```
violetagg, line 49: I’m not sure 
I think it should be 1.4.0-SNAPSHOT

violetagg, line 49: I think that for the stable release we have to use `1.3.23` as we do not have any other, while for the snapshot we have to use `1.4.0-SNAPSHOT`

```suggestion
      // compile "io.projectreactor.kafka:reactor-kafka:1.4.0-SNAPSHOT"
      compile "io.projectreactor.kafka:reactor-kafka:1.3.23"
```

chemicL: I based this on https://github.com/reactor/reactor-kafka/commit/3877079c63d36ae690f9911a6ac8808d5d14522d

I can change if you think that makes more sense.

violetagg: Yeah until we have `1.4.0-M1` we don't have any other stable release.
Once we have `1.4.0-M1`, I would expect to see:

```
// compile "io.projectreactor.kafka:reactor-kafka:1.4.0-SNAPSHOT"
compile "io.projectreactor.kafka:reactor-kafka:1.4.0-M1"
```

See https://github.com/reactor/reactor-core/tree/v3.6.0-M1

PR #390, opened 2024-06-05
Migrate reference documentation to Antora

pderop:
This PR attempts to migrate Reactor Kafka reference documentation to Antora, and includes the following logical commits:

- 492f4075b4bdbf95c0a1ad379e0cae825b57c1f9: moved adoc files to Antora directory structure
- ee74aeafcc1fffcda99b5f5cf9758d729bb9ca62: added source highligher
- a2c98ec5ab28edbbc82c51988ec5a8f6dd425bd0: adapted adoc files to Antora
- 20becb55dcc621b878cce8a5d6b83ef75879b05c: updated gradle version to be able to build antora docs using jdk version >= 17
- 8a28b2de211c8fdfc02fdb6b40c6d1c45e9527cf: added antora yaml files
- b27e6752e0e84191e3ab6f8958d7361568099463: adapted gradle scripts for antora
- 6e0e8b575b198cf1be9b1e07c5d6dc6cf9269fa2: updated github publish workflow in order to build docs using JDK17+ version separately

### How to build and publish the doc to your local M2:

You must install two JDKs in your system: JDK8 (used to build and test everything, as before), and a JDK17 compatible JDK version (>= 17) that is required to build the antora docs.

if you only install JDK8, the documentation won't be generated and won't be included in your local M2 when doing for example `./gradlew publishToMavenLocal`).

Now, there are two ways to build and publish the doc: in two steps or in one shot.

#### Building or publishing the doc in one shot:

Set the current active JDK to the one that is compatible with JDK17, then type usual commands to build and/or deploy the docs (`./gradlew docs` or `./gradlews publishToMavenLocal`)

if you need to also include PDF, then add `-PforcePdf` option in the gradle options.

the JDK17+ version is only used to generate the doc, everything else is built using your other JDK8 version.

#### Building or publishing the doc in two steps:

if you need to build reactor-kafka using JDK8, then you need to build the doc in two steps:

- step 1: set the current active JDK to the one that is compatible to JDK17+ version, then generate the doc with `./gadlew  docs` or `./gradlew docs -PforcePdf`

- step 2: set the current active JDK version to the JDK8 JDK, then when you type `./gradlew publishToMavenLocal` or `./gradlew publishToMavenLocal -PforcePdf`, the reactor-kafka-<version>-SNAPSHOT-docs.zip artifact will also be published in your local M2 (it will include the docs generated in previous step 1).

**Notes**: 

- if you only install JDK8, doing `./gradlew docs` won't work, and `./gradlew publishToMavenLocal` won't include the reactor-kafka-<version>-SNAPSHOT-docs.zip file.

- when also generating PDF documentation, you first need to pre install the asciidoctor-pdf tool because the antora pdf assembler requires it:

```
brew install asciidoctor
```


pderop:
@violetagg , or @chemicL ,

can someone of you please have a look ?

PR #402, opened 2024-10-24
Regroup Dependabot updates

artembilan:
* Kafka and Micrometer deps as `type/dependency-upgrade`
* All other as `type/chores`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,7 +26,6 @@ updates:
       day: sunday
     ignore:
       - dependency-name: '*'
-    open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


artembilan:
Also rename label to `type/chores` to realign with whatever is expected by change log generator in Project Reactor: https://github.com/reactor/.github/blob/main/.github/release.yml#L22

violetagg (approved):

artembilan:
Or we can try like that.
I'm not sure in anything 😄 
So, if you want I can merge and go from there.
Thanks

violetagg (approved):

PR #428, opened 2025-11-06
Fix usage of deprecated ConsumerRecord constructor

fstaudt:
ReceiverRecord still uses a deprecated constructor of `org.apache.kafka.clients.consumer.ConsumerRecord` (deprecated since 3.0).\
It prevents to use reactor-kafka with 4.x versions of Kafka library (e.g. 4.1.0 provided by Spring Boot BOM 4.0.0-RC1):

```
java.lang.NoSuchMethodError: 'void org.apache.kafka.clients.consumer.ConsumerRecord.<init>(java.lang.String, int, long, long, org.apache.kafka.common.record.TimestampType, java.lang.Long, int, int, java.lang.Object, java.lang.Object, org.apache.kafka.common.header.Headers)'
	at reactor.kafka.receiver.ReceiverRecord.<init>(ReceiverRecord.java:48)
```

The deprecation can be avoided by using constructor of ConsumerRecord used under the hood by the deprecated constructor.\
It also simplifies the code of ReceiverRecord.

I acknowledge that this fix will not guarantee compatibility of reactor-kafka 1.3.x with Kafka 4.x library or Spring Boot 4.0.0.\
It will however fix a known compatibility issue and simplify maintenance of ReceiverRecord.

### DIFF
--- a/src/main/java/reactor/kafka/receiver/ReceiverRecord.java
+++ b/src/main/java/reactor/kafka/receiver/ReceiverRecord.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2025 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


chemicL (approved):

chemicL:
Thank you @fstaudt, it looks that we can merge this. Can you please update the license headers as per the build output using `./gradlew spotlessApply`?

fstaudt:
> Thank you @fstaudt, it looks that we can merge this. Can you please update the license headers as per the build output using `./gradlew spotlessApply`?

Thanks for the quick reply.\
I've used spotlessApply to update license headers on the updated file.

Kr,
François

